PR #20, opened 2015-01-31
Add support for .marko.html templates

bkniffler:
There are several use-cases where you’d want .html files instead of .marko files. So I’ve added support for „.marko.html“ files.

### DIFF
--- a/bin/markoc.js
+++ b/bin/markoc.js
@@ -242,7 +242,7 @@ if (args.clean) {
             file: function(file, context) {
                 var basename = nodePath.basename(file);
 
-                if (basename.endsWith('.marko.js') || basename.endsWith('.marko.xml.js')) {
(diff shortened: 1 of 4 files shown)


patrick-steele-idem:
I have no objections to adding support for `*.marko.html` except that most editors only let you set file associates based on the last part. Therefore, you would likely lose the ability to have custom syntax highlighting for Marko template files and instead have to settle for HTML syntax highlighting, which is not too terrible I guess.

Please see my comment I added to your Pull Request and please make sure the tests pass by running `npm test`. 

Thanks for sending the PR.

bkniffler:
Hey, thanks for the quick reply, I've refactored the code as requested, it should be all fine now. The tests do pass.

patrick-steele-idem:
New version published: `marko@1.4.0`

patrick-steele-idem:
Thanks for the PR, @bkniffler

PR #30, opened 2015-02-05
Support $global in template.renderSync

ianma:
* https://github.com/raptorjs/marko/issues/28


patrick-steele-idem:
Thanks for the fix! New version published: `marko@1.5.4`

PR #43, opened 2015-03-03
marko with webpack

maxwarch:
Hello,

is there a way to use marko with webpack ?

Thanks


patrick-steele-idem:
@maxwarch I think you meant to ask this question as a Github issue and not a Pull Request. Please open a Github issue with your question and we can discuss there. Thanks.

PR #49, opened 2015-03-18
Allowing complex var names;

seangates:
Removing the else if block in order to allow complex var names on the left hand of the assignment.


philidem:
Please fix the jshint error so that the build will pass. Also, jshint plugins/packages for your code editor are really helpful if you don't have one installed.

See https://travis-ci.org/raptorjs/marko/jobs/54955691

seangates:
Funny, because I did the edits directly in Github. :-)

--
Sean Gates

philidem:
Published marko@2.1.4
Thanks Sean!

seangates:
You bet!

--
Sean Gates

PR #62, opened 2015-04-10
Fixes #53 Merge c-input with attr props

### LINKED ISSUE #53: Support sending both c-input and attribute for custom tags
`c-input` works well for nested tags, but there are times when we need to pass down extra information from the parent tag, like this:

```
<tagname c-input="data.someObj" attribute="test"/>
```

Examples include when the parent contains the loop and we might need an index in the nested tag, or if parent tag-level logic affects a child tag. 

This should also include support for passing a conditional, or at least a variable (that we can set with a conditional earlier in the file).

Thanks!

patrick-steele-idem:


philidem:
Can you include a snippet of the compiled output for us to review? It looks like you're using `extend` (or the `xt` helper for short) which seems like a good approach.

patrick-steele-idem:
Ok, I'll add better code comments, but the following is the generated code:

```javascript
__tag(out,
      simple_hello_tag,
      __extend(__extend({}, {name: 'Jane', adult: true}), {
        "name": "Frank"
      }));
```

There are two extends because we don't want to modify the original object. The inner extend creates a new object with the properties of the `c-input` expression merged in. The outer extend extends the inner object with the properties from the attributes.

taglibs/core/TagHandlerNode.js
```diff
@@ -177,17 +178,42 @@ TagHandlerNode.prototype = {
...
+                        // the provided data object. We don't want to modify the existing
+                        // data object provided by the user so first need to create a new
+                        // empty object and then merge in the existing properties from the
+                        // provided object. When then extend that object with the properties
+                        // that came from the attributes.
+                        //
+                        // The generated code will be similar to the following:
+                        //
+                        // extend(extend({}, <input_expression>), <attr_props>);
+                        var extendVar = template.addStaticVar('__extend', '__helpers.xt');
```
ramahadevan: How will the extendVar look like?

patrick-steele-idem: The `template.addStaticVar('__extend', '__helpers.xt');` line will result in a new JavaScript variable introduced in the template with the following code:

```javascript
var __extend = __helpers.xt;
```

The value of `extendVar` will be the name of the variable (i.e. `"__extend"`).

ramahadevan:
Looking good

PR #67, opened 2015-04-23
Fixes #66 - Allow circular dependencies when loading templates

### LINKED ISSUE #66: Circular dependencies between tags/templates causes problems (Maximum call stack size exceeded)

patrick-steele-idem:


patrick-steele-idem:
This Pull Request makes a slight change to how Template instances are loaded and created to avoid problems with circular dependencies. The partially loaded Template instance is immediately put into a cache before being fully initialized. In the case of a circular dependency the partially loaded Template instance will be returned, but that is okay because it will eventually be fully loaded.

runtime/helpers.js
```diff
@@ -22,6 +22,26 @@ function notEmpty(o) {
     return true;
 }
 
+function createLazyRenderer(handler) {
```
ramahadevan: Is it really a lazyRenderer ? seem like the handler.renderer is abstracted here ..

patrick-steele-idem: I agree that we can probably improve the name here. Maybe `createDeferredRenderer` is better? If you don't like that, maybe you can come up with a better name? When there are circular dependencies we may end up with a partially loaded module that does not yet have it's `renderer(input, out)` method. When we detect that situation we create a wrapper renderer function that on first render tries again to resolve the actual `renderer` function on the provided module. I think `defer` is better because it means "put off (an action or event) to a later time; postpone.". Thoughts?

ramahadevan: Thats fine.. this is trivial, the name implied a lazy loading.. I will merge it

PR #71, opened 2015-05-03
update README.md: tags-dir array support

aselvaraj:


patrick-steele-idem:
Thanks for improving the documentation!

PR #77, opened 2015-05-21
further tighten entity-matching regex, add basic test for entity matching

DanCech:
This tightens the entity matching regular expression to explicitly handle numeric and hexadecimal entities separately from named entities.  It also adds a simple test that entities are being properly matched.


patrick-steele-idem:
Thanks for the improved fix!

patrick-steele-idem:
New version published with fix: `marko@2.4.3`

PR #79, opened 2015-05-22
Fixes #78 Require extension for .marko files

### LINKED ISSUE #78: Proposal: allow require('./template.marko')
The following code works, but is a lot of code to load a template:

```javascript
var template = require('marko').load(require.resolve('./template.marko'));
```

We can do better:

```javascript
var template = require('./template.marko');
```

For this to work, some code must register a custom Node.js require extension handler for `.marko` files using the proposed code:

```javascript
require('marko/node-require').install();
``` 

Internally, the above code will do something similar to the following:

```javascript
require.extensions['.marko'] = function(module, filename) {
        // Compile the Marko template to JavaScript and load the
        // JavaScript source as a Node.js module
};
```

# Q&A

__Question: The usage of [`require.extensions` is deprecated](https://nodejs.org/api/globals.html#globals_require_extensions). Isn't it a bad idea to rely on a deprecated feature?__

Answer: That functionality will very likely never be able to removed even though it is deprecated. Personally, I feel like deprecating that feature without a suitable alternative was misguided. In our case, we will solve the "problem" of a global require extension by having the global require extension simple delegate to the correct version of the module (see later question).

-----------------

__Question: What happens if the code to register the `.marko` require extension is called multiple times?__

Answer: Only the first call will register the require extension. Additional calls will do nothing.

-----------------

__Question: What if there are different versions of Marko installed in the application and they are all using a globally installed require extension?__

Answer: The require extension will resolve the `marko` module relative to the template being required. That is, the globally installed require extension for `.marko` files will delegate the compiling and loading of the Marko template to the correct version of the marko module. It won't matter which code actually registers the `.marko` require extension!

-----------------

__Question: Will backwards compatibility be maintained?__

Answer: Yes. Old code will continue to work exactly the same.

patrick-steele-idem:


README.md
```diff
@@ -298,12 +318,38 @@ Finally, another distinguishing feature of Marko is that it supports _asynchrono
...
+```javascript
+// The following line installs the Node.js require extension
+// for `.marko` files. Once installed, `*.marko` files can be
+// required just like any other JavaScript modules.
+require('marko/node-require').install();
+
+// ...
+
+// Load a Marko template by requiring a .marko file:
+var template = require('./template.marko');
```
dimichgh: Will this work in browser?

README.md
```diff
@@ -298,12 +318,38 @@ Finally, another distinguishing feature of Marko is that it supports _asynchrono
...
+
+A loaded Marko template has multiple methods for rendering the template as described in the next section.
+
 ## Template Rendering
 
 ### Callback API
 
 ```javascript
-var template = require('marko').load('template.marko');
+var template = require('./template.marko');
```
dimichgh: in case I need to cleanup cache, how do I do it?

README.md
patrick-steele-idem: I just pushed another round of commits that change compiled templates to export a loaded `Template` instance. In the previous version of marko, compiled templates exported a "create" function that could be used to create a loaded `Template` instance. Now that compiled templates export a loaded `Template` instance, the compiled template module can be required to get a loaded `Template` instance and this will work the same on the server and in the browser (as long as the `template.marko.js` file gets transported and used as the require). Backwards compatibility has been maintained so that if a template was compiled with a previous version of marko then it will continue to be loaded just fine (tests were added).

README.md
patrick-steele-idem: In the latest version of marko the only cache is the Node.js require cache (i.e. `require.cache`). In order to reload a template without restarting the server it is necessary to delete the entry out of the require cache. This is what the hot-reload module does (see https://github.com/patrick-steele-idem/marko/blob/7c00085380ea318a9b9a935760e5e323d12612b7/hot-reload/index.js#L92-L93)

PR #101, opened 2015-07-11
Invalidate cache before requiring changed template

bard:
`marko.load` checks if template is up to date and, if not, it compiles,
saves, and requires it again. However, it saves it under the same name
as the old template, so just requiring it again hits the module cache
and returns the compiled template.

Fix by deleting old template from module cache before re-requiring.

Failing test case:

    var fs = require('fs'),
        marko = require('marko'),
        data = { what: 'world' };

    fs.writeFileSync('test.marko', 'hello, ${data.what}!');
    marko.load('test.marko').render(data, function(err, out) {
        console.log(out); // prints "hello, world"

        fs.writeFileSync('test.marko', 'goodbye, ${data.what}!');
        marko.load('test.marko').render(data, function(err, out) {
            console.log(out); // prints "hello, world" again
        });
    });

### DIFF
--- a/runtime/loader.js
+++ b/runtime/loader.js
@@ -53,6 +53,7 @@ module.exports = function load(templatePath) {
     fs.writeFileSync(tempFile, compiledSrc, fsReadOptions);
     fs.renameSync(tempFile, targetFile);
 
+    delete require.cache[require.resolve(targetFile)];
     return require(targetFile);
 };
 
--- a/runtime/loader.js
+++ b/runtime/loader.js
@@ -53,7 +53,7 @@ module.exports = function load(templatePath) {
     fs.writeFileSync(tempFile, compiledSrc, fsReadOptions);
     fs.renameSync(tempFile, targetFile);
 
-    delete require.cache[require.resolve(targetFile)];
+    delete require.cache[targetFile];
     return require(targetFile);
 };
 


patrick-steele-idem:
Hi @bard,  thanks for the Pull Request, but the code in `runtime/loader.js` is not meant to handle hot reloading of files. The hot reloading functionality enables a template at template to be reloaded at runtime without restarting the app. Hot reloading is handled separately by the following module:
https://github.com/raptorjs/marko/blob/master/hot-reload/index.js

The hot reloading code includes code to delete the loaded template out of the require cache:
https://github.com/raptorjs/marko/blob/4596c936684a669ac2cbfea82008eca8a2ad38a4/hot-reload/index.js#L92-L93

Here's some documentation and a sample project that illustrates how to enable hot reloading of templates:
https://github.com/marko-js-samples/marko-hot-reload

Also, we have some test cases to ensure that hot reloading works that might make things a little clearer:
https://github.com/raptorjs/marko/blob/master/test/hot-reload-test.js

Since the tests are failing with your changes I am going to close this PR, but please let me know if you are still unable to get hot reloading working in your project and I can help out.

Thanks again.

bard:
Thanks for the explanation and for the workaround!

I'm using marko in a gulp pipeline to generate a static web site. This issue came up when I switched from invoking gulp manually to having it watch files and re-run the pipeline when changes are detected. Obviously in the latter case all runs after the first are on a hot `require` cache.

I tried the hot-reload mechanism you suggested and it works fine.

I'd still like to suggest that there's a code smell: `marko.load(templatePath).render(...)` sounds to me like a manual/forced reload rather than a hot reload. Note that both in the tests and the sample you linked to there's no manual invocation of `load()`, so indeed hot reload.

That `load()` recompiles the template if it's outdated seemed to support the idea of a manual/forced reload. After each run I was checking compiled templates and finding them up to date, which left me puzzled as to why they weren't being taken into account. That `load()` recompiles but doesn't reload, then, struck me as inconsistent.

I guess it comes down to the question of what one expects by reading code like this and whether that matches what is happening:

    marko.load(templatePath).render(data);
    modifyTemplate(templatePath);
    marko.load(templatePath).render(data);

If others read it as I did, i.e. that the third line will produce a different result from the first, they may have some code digging to do. (Hopefully they'll check closed issues and stumble on this. :))

Just .02$!

PR #115, opened 2015-08-05
doc reference for getLength() of loop status-var

yomed:
The utility function `getLength()` for loop status variables was previously undocumented.


patrick-steele-idem:
Nice improvement. Thank you!

PR #133, opened 2015-08-25
add tag-body attribute to render tag body using invoke renderBody

kristianmandrup:
It works just fine, except html equality test breaks even though I copy paste `actual.html` and `expected.html` to match!?. 

See `test/fixtures/templates/tag-body`


patrick-steele-idem:
Hi @kristianmandrup, I'm not sure if it makes sense to pull this into core just yet. Here's the same functionality provided as a separate module published to npm: https://github.com/marko-js/marko-tag-body

Does that work for you?

kristianmandrup:
Awesome! Much better :) Thanks a lot! 

Of course this solution is just excellent! I learned a lot from my experiments, and this little helper lib you made clearly demonstrates how to enhance Marko in an elegant way!!! 

Have a nice evening @patrick-steele-idem  :)

patrick-steele-idem:
There's some really cool stuff that you can do simply by transforming the AST during the compilation stage. It definitely needs to be documented. I'm glad you were able to get your hands dirty and learn some of the tricks for marko!

kristianmandrup:
Cool :) Yeah, next step IMO is to improve the documentation some more. 
Perhaps by going away from having everything in one huge README? A github "blog/page" or wiki?
I plan to add a `tag-body-or` convenience helper today for the very common case of rendering body or attribute as body.

kristianmandrup:
On another note, I was wondering if there is a way to dynamically generate the tag name, such that:

```html
h1(class='ui header' tag-body='data.label' if='data.lv == 1')
h2(class='ui header' tag-body='data.label' if='data.lv == 2')
h3(class='ui header' tag-body='data.label' if='data.lv > 2')
```

Could be rewritten something like

```html
<dynamic-tag name='h${data.lv}' class='ui header' tag-body='data.label'/>
```

Would love to have another small lib like the one you created for this use case :)

kristianmandrup:
I would think I could achieve this something like this?

```js
module.exports = function transform(node, compiler, template) {
    var name = node.getAttribute('tag-name');
    if (name != null) {
        node.removeAttribute('tag-name');
        var tagNode = this.compiler.createNode(name, node.getAttributes());

        node.parentNode.replaceChild(tagNode, node);
        node.forEachChild(function(child) {
          tagNode.appendChild(child);
        })
    }
};
```

kristianmandrup:
Fuck, I spent all morning trying to figure out how to evaluate an expression :P
Like trying to find a needle in a haystack...

PR #136, opened 2015-08-28
add ElementNode.generateTagName function 

kristianmandrup:
to allow for dynamic runtime tag name transformation :)

This will enable my https://github.com/kristianmandrup/marko-dynamic-tag to work :)
And loads of other powerful use cases...!!!

### DIFF
--- a/compiler/ElementNode.js
+++ b/compiler/ElementNode.js
@@ -89,9 +89,18 @@ ElementNode.prototype = {
 
         attributes.forEach(callback, thisObj);
     },
+    setSkipAttributes: function(attNames) {
+      this.attributesToSkip = attNames;
+    },
+    addSkipAttribute: function(attName) {
+      this.attributesToSkip = this.attributesToSkip || [];
+      this.attributesToSkip.push(attName);
+    },
+    isSkipAttribute: function(attr) {
+      return this.attributesToSkip.indexOf(attr.localName) >= 0;
+    },
     forEachAttributeNS: function (namespace, callback, thisObj) {
         namespace = this.resolveNamespace(namespace);
-
         var attrs = this.attributesByNS[namespace];
         if (attrs) {
             forEachEntry(attrs, function (name, attr) {
@@ -204,6 +213,7 @@ ElementNode.prototype = {
     },
     // override for optional tag name transformation
     generateTagName: function (name, template) {
+      console.log('ElementNode generateTagName', name);
       template.text(name);
     },
 
@@ -220,6 +230,8 @@ ElementNode.prototype = {
         this.generateTagName(name, template);
 
         this.forEachAttributeAnyNS(function (attr) {
+            if (this.isSkipAttribute(attr)) return;
+
             var prefix = attr.prefix;
             if (!prefix && attr.namespace) {
                 prefix = this.resolveNamespacePrefix(attr.namespace);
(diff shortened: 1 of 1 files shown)


patrick-steele-idem:
Hi @kristianmandrup, thanks for the PR, but your changes broke tests (please run `npm test` first). I hope you don't mind, but I went ahead and made changes in marko core to support the following:

```xml
<html-element tag-name="hello-${data.myTagName}" class="my-class" foo="bar">
    My nested content
</html-element>
```

Output:

```html
<hello-world class="my-class" foo="bar">
    My nested content
</hello-world>
```

See: https://github.com/marko-js/marko/issues/137

kristianmandrup:
Awesome :) I sure don't mind!!!

kristianmandrup:
Would it also be possible to dynamically decide which component (function I assume) to call using a similar approach?

kristianmandrup:
BTW: On the PR, all the tests passed on my local machine running `mocha`, so I was surprised to see it fail on Travis CI.

patrick-steele-idem:
Not exactly... supporting dynamic components is a little tricky. It's kind of support if you use `<invoke function="..."/>` but that's not a clean solution. We talked about adding support for something similar to the following:

```xml
<register-dynamic-tag tag="ui-component-one"/>
<register-dynamic-tag tag="ui-component-two"/>

<dynamic-tag tag="ui-component-${data.variant}" name="Frank" message-count="${30}"/>
``` 

Is that along the lines of what you are thinking?

kristianmandrup:
Exactly what I was thinking!!! :)

kristianmandrup:
In my scenario I would have an incoming list with varying types of data (ie. some attributes undefined)...
Depending on the data I what to render an appropriate tag, such as icon-item if it has an icon, link-item if there is a link and so on. Of course I could have a "catch all" item, but it would have to contain loads of conditional logic... the dynamic tag would come in very handy to make for a more elegant solution :)
Cheers!

patrick-steele-idem:
It's doable, but it would require a little more work than what was required for `<html-element>`. I opened a Github issue: https://github.com/marko-js/marko/issues/139

I support that proposal, but I won't be able to work on it for at least a week or two. I need to finish up the marko website ([http://markojs.com.com](http://markojs.com/)!!!) and put together a blog post announcing Marko Widgets v5.

In the meantime, I recommend:

```xml
<app-foo name="Frank" message-count="${30}" if="data.variant === 'foo'"/>
<app-foo name="Frank" message-count="${30}" else-if="data.variant === 'bar'"/>
<app-baz name="Frank" message-count="${30}" else/>
```

kristianmandrup:
An example:

Here I decide whether to render a checkbox as a `toggle`, `slider`, or a normal `checkbox`. 
Not very elegant, but gets the work done :P

```jade
ui-toggle(attrs='data' if='data.toggle')
ui-slider(attrs='data' if='data.slider')
div(class='$data.field field' body-only-if='!data.field' if='data.default')
  div(class='ui $data.ui checkbox' body-only-if='!data.ui')
    input.hidden(type='checkbox' name='$data.name' checked='$data.checked' tabindex='$data.tab')
    label(if='data.label') $data.label
```

kristianmandrup:
Sweet :) Thanks for the tip. Didn't even know you could do complex conditional logic inline like that :)
I agree, those tasks are way more important for now than these obscure bonus features!!!
Good job :) Looking better every day and every week :D

PR #144, opened 2015-09-11
template.stream() is 'error' emitter

jlarsson:
Marko + koa currently misbehaves due to errors thrown during template stream rendering are uncaught, leading to hard crashes of node.

The sample code below illustrates a very simplied setup, and this patch ensures that errors during template rendering are indeed picked up by the koa error handling system.
```
// Simplified koa usage
app.use(function * () {
  this.type = 'text/plain'
  this.body = require('template.marko').stream({...})
})

app.on('error', function (err) {
  log.error(err) 
})
```

Thanks for an awesome piece of work. Marko+koa is my bread and butter...

### DIFF
--- a/runtime/marko-runtime.js
+++ b/runtime/marko-runtime.js
@@ -189,7 +189,17 @@ Template.prototype = {
 
         // Invoke the compiled template's render function to have it
         // write out strings to the provided out.
-        renderFunc(finalData, finalOut);
+        try {
+          renderFunc(finalData, finalOut);
+        }
+        catch (e) {
+          if (typeof finalOut.on === 'function') {
+            finalOut.emit('error', e);
+          }
+          else {
+            throw e;
+          }
+        }
 
         // Automatically end output stream (the writer) if we
         // had to create an async writer (which might happen
--- a/test/api-tests.js
+++ b/test/api-tests.js
@@ -336,4 +336,15 @@ describe('marko/api' , function() {
             stream);
     });
 
+    it('should emit errors when rendered with stream', function(done) {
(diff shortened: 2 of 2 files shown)


patrick-steele-idem:
Hi @jlarsson, thanks for the PR. I see where you are going with this but I do have some concerns. Adding a `try...catch` will prevent the frequently called code from being optimized by V8 and it will also only catch any synchronous errors. I'm more inclined to add the `try...catch` block around the following line:

https://github.com/jlarsson/marko/blob/3c51134965474d30de7e4e7e82f68cb3de822d0b/runtime/marko-runtime.js#L252

Thoughts? Can you make that change?

On a related note, Koa relies on a returned `Readable` stream, but in most Express apps we are writing directly to the `Writable` HTTP response stream without creating an intermediate stream (faster...):

```javascript
var template = require('./template.marko');

module.exports = function(req, res) {
    template.render({ ... }, res);
}
```

 Just throwing that out there :)

Glad you are enjoying marko!

jlarsson:
@patrick-steele-idem, I get your points regarding V8. 
The more idiomatic use of ```render()``` works well with Koa too, as in
```
app.use(function * () {
  template.render({...}, this.res)
})  
```
A simple change for me and I accept that as an answer :)

I think I have spotted an edge case here. Everything that goes down to Async-Writer/async-fragment seems to work super nice, It's just the initial synchronous invocation that needs some loving care. 
I'm still rather dependent on streams in some peculiar situations, so I will try to make a better PR respecting performance and also covering asynchronous situations.

Feel free to discard this PR ;)

PR #145, opened 2015-09-15
conditional unless added

kristianmandrup:
Please review and merge :)


kristianmandrup:
On my laptop I ran `mocha` and had   `181 passing (11s)` (all tests passing)

patrick-steele-idem:
Thanks for the PR. 

I see the following jshint error that looks like a problem:

```
taglibs/core/core-tag-transformer.js: line 113, col 55, 'UnlessNode' is not defined.
```

Can you please add tests and update the docs? For the docs you will need to update the following file:
https://github.com/marko-js/marko/blob/master/docs/language-guide.md

For the tests, please duplicate the following directory and update it to test the new `unless` attribute and the new `<unless>` tag:
https://github.com/marko-js/marko/tree/master/test/fixtures/templates/if-else

Thanks,
Patrick

kristianmandrup:
Alright, will do :)

kristianmandrup:
Whenever I use expected and actual, I always end up in this strange situation:

```sh
  1) render unless-else:
     Unexpected output for "/Users/kristianmandrup/repos/markoa/marko/test/fixtures/templates/unless-else/template.marko":
EXPECTED (/Users/kristianmandrup/repos/markoa/marko/test/fixtures/templates/unless-else/expected.html):
---------
B , A , A , <div>A</div>

---------
ACTUAL (/Users/kristianmandrup/repos/markoa/marko/test/fixtures/templates/unless-else/actual.html):
---------
B , A , A , <div>A</div>
```

I even copy paste the actual into expected, and still the same error! Must be a whitespace issue?

kristianmandrup:
I also get this warning for `if-else`

```sh
    ✓ if-else
WARN marko/runtime/helpers: data.invokeBody() deprecated. Use data.body instead. Error
    at Object.data.invokeBody (/Users/kristianmandrup/repos/markoa/marko/runtime/helpers.js:326:89)
```

patrick-steele-idem:
We are using a simple string compare for "actual" and "expected" and any whitespace difference will cause the test case to fail. I suspect you are using Atom and it is automatically adding a trailing new line when you save the "expected" file to disk. You can disable the feature by adding the following to your `User Config` file:

```
"*":
  // ...
  whitespace:
    ensureSingleTrailingNewline: false
```

kristianmandrup:
Yup exactly! Using Atom, haha :P Thanks for the tip. Will try again!

patrick-steele-idem:
Hi @kristianmandrup, the code changes and test look good, but why did you completely reformat the markdown for the docs? It looks like the markdown changes were just stylistic changes, but now the style will be different in the other markdown files. I would rather use the old style unless there is a good reason to switch. Can you please only change the relevant docs for `unless` and we can discuss stylistic changes to the Markdown as a separate Github issue/PR. Thanks.

kristianmandrup:
Really, hmm... that was not my intention. Guess I made a Markdown error of some sort. 
I only inserted a few more examples in the relevant section. I'll fix that shortly.

kristianmandrup:
I just reviewed my changes and reverted what I think you call "stylistic changes". If you are still not happy, please edit as you like. I tried my best to improve the docs while keeping with the original intent and styling.

patrick-steele-idem:
Thanks for the PR, @kristianmandrup. New version published: `marko@2.7.22`

The stylistic changes that I was referring to was the Markdown syntax for declaring headings and bold. I reverted those changes and updates the docs for Conditionals:
- https://github.com/marko-js/marko/commit/4a58db78ee36acd39bc90a18bee24578863e904d
- https://github.com/marko-js/marko/commit/1e19886a5a443f1b4ad6d5b90fba89d9dfc8fd61

I also published the new docs to markojs.com. Please let me know if you see any issues. Thanks again.

PR #147, opened 2015-09-18
Fixed a typo for rendererFunc in helpers.js

pdcopeland727:
This should resolve issues with circular dependency rendering. Previously, whenever this function was called, it would throw an error because `typeof renderFunc` was always `undefined`. With this fix, `rendererFunc` will (in valid cases) be a function.


patrick-steele-idem:
Thanks for the fix! New version published: `marko@2.7.23`

PR #153, opened 2015-10-06
Make loading template from String template source easier

philidem:
- Allow loading of template from template source on server-side
- Fix documentation for JavaScript API
- Allow `writeToDisk` option to be passed to load function (e.g. `require('marko').load(templatePath, {writeToDisk: false})`)


patrick-steele-idem:
New version published: `marko@2.7.27`

Thanks for the improvement.

PR #156, opened 2015-10-16
Allow tag name to be set in renderer

zephraph:
It might not be as simple as this, but one can hope, right?

The purpose of this change is to allow this to be written in the renderer

```javascript
module.exports.tag = { name: "my-tag-name" }
```

That would result in the template being named to whatever was specified.

### DIFF
--- a/compiler/taglibs/taglib-loader/scanTagsDir.js
+++ b/compiler/taglibs/taglib-loader/scanTagsDir.js
@@ -137,13 +137,13 @@ module.exports = function scanTagsDir(tagsConfigPath, tagsConfigDirname, dir, ta
 
                 tagDef.renderer  = rendererJSFile;
                 tag = loader.tagLoader.loadTag(tagDef, tagsConfigPath, taglib, tagDirname);
-                tag.name = tag.name || tagName
+                tag.name = tag.name || tagName;
                 taglib.addTag(tag);
             }
 
             if (tagDef) {
                 tag = loader.tagLoader.loadTag(tagDef, tagsConfigPath, taglib, tagDirname);
-                tag.name = tag.name || tagName
+                tag.name = tag.name || tagName;
                 taglib.addTag(tag);
             }
         }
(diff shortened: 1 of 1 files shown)


philidem:
You'll need to make sure tag.name is not overridden due to this line:
https://github.com/marko-js/marko/blob/master/compiler/taglibs/taglib-loader/scanTagsDir.js#L140

zephraph:
So I'm looking for the proper location to place the renderer fixture that I use for the test. Should it just go in `test/fixtures/taglib`? Also, any preferred name for it?

patrick-steele-idem:
Hey @zephraph, are you still interested in this PR? My opinion is that I don't think it is a good idea for the tag name to be defined in the renderer or in the `marko-tag.json`. I think it is better for the tag name to be defined in the `marko-taglib.json` file or have it be set based on the directory name (when using the directory scanning feature). What do you think?

zephraph:
I definitely still think it has value. I try to put as little as possible in `marko-taglib.json` or any `marko-tag.json`. With that being said I think basing it on the folder name that it's inside is a valid compromise. 

Ultimately if you keep the ability to set it in the `marko-tag` I think this PR should happen. If not, then it likely shouldn't. 

It's just nice to me:

```javascript
export default defineComponent({
  name: 'gh-notification',
  template,
  ...
});
```

patrick-steele-idem:
I'm also a fan of minimizing how much code gets put into configuration files such as `marko-tag.json` and `marko-taglib.json` so I think it is better to have the convention of using a directory scanner to discover custom tags and for the custom tag name to automatically match the name of the parent directory. I think it would be a mistake to allow the developer to choose a tag name that doesn't match the parent directory name since it would make things confusing.

Also, currently, if you _aren't_ using a directory scanner then you must use a _map_ to define custom tags and that requires that the tag name always be declared in the `marko-taglib.json`:

```json
{
  "<my-custom-tag>": "./path/to/my-custom-tag/marko-tag.json",
  "<another-custom-tag>": { "@foo": "string", ... }
}
```

It wouldn't make sense to allow the developer to redefine the tag name in the `marko-tag.json` or in the `renderer.js` file since it was already declared in the parent taglib.

I propose that we not merge this PR if you are cool with that :) Let me know.

zephraph:
Sounds good.

PR #157, opened 2015-10-16
Use new travis infrastructure

zephraph:
Travis upgraded its infrastructure which allows CI tests to be ran in docker containers. The result is faster tests. 

To use it you just have to specify `sudo: false` in the travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 node_js:
 - "0.10"
 - "0.12"


philidem:
Seems reasonable! Double-checked the travis documentation and this looks like a safe way to force container infrastructure to be used.

patrick-steele-idem:
Thanks for the fix, @zephraph.

I don't think your change was the cause, but for some reason the Travis CI build is failing now :disappointed: I'll retry in a bit, but if you find a solution please share.

zephraph:
@patrick-steele-idem which build? The only thing that I saw that has failed recently was a rebuild that triggered on my PR after it was merged (which failed because the git ref that PRs point to are deleted after merge).

philidem:
I think the failed build was my fault because closed the pull request and
then reopened it. The build error message seem to imply that this was the
reason for failure. Subsequent builds have succeeded.

PR #160, opened 2015-11-09
marko.stream -> marko.load..stream

designeng:


patrick-steele-idem:
Hey @designeng, thanks for the PR, but I think we need to keep the test case as-is. While `require('marko').stream(templatePath)` is not the recommended API, it is still supported and we should keep that test case around to make sure it continues to work. There is another test case that tests `marko.load..stream': https://github.com/marko-js/marko/blob/56d23b07bdc4ca2a20ebce868a3982277c97c618/test/api-tests.js#L170-L191

Look good to you?

designeng:
Hey @patrick-steele-idem, I see, absolutely!
Thank you for brilliant template engine and very interesting test harness created for it.

patrick-steele-idem:
Thanks, @designeng! I appreciate the feedback.

Issue #169, opened 2015-12-08
Marko v3: var tag


patrick-steele-idem:
Single unscoped variable:

```xml
<var x=1 />
```

Output JavaScript code:

```javascript
var x=1;
```

Single unscoped variable with no init value:

```xml
<var x />
```

Output JavaScript code:

```javascript
var x;
```

Multiple unscoped variables:

```xml
<var x=1 y=7 z=x+10 unusedVar/>
```

Output JavaScript code:

```javascript
var x=1,
    y=7,
    z=x+10,
    unusedVar;
```

Scoped variables:

```xml
<var x=1 y=7 z=x+10 unusedVar>
  ${x}
  ${y}
  ${z}
</var>
${x} <!-- Runtime error! Uncaught ReferenceError: x is not defined -->
```

Output JavaScript code when using scoped variables:

```javascript
(function() {
    var x=1,
        y=7,
        z=x+10,
        unusedVar;
   ...
}())
```

__Important:__ As part of a more powerful `<var>` tag we should drop support for the `<with>` tag and the `@with` attribute!

BryceEWatson:
Hi @patrick-steele-idem, I have two questions:

1) I noticed that the [pending test for var](https://github.com/marko-js/marko/blob/htmljs-parser/test/fixtures/render/autotest-pending/var/template.marko) has the following example in it:
```
<var name="colors" value="['red', 'green', 'blue']"/>
<div for="color in colors">
    $color
</div>
```
Do we want to include this logic, where if there is both a name and value then we want to generate `var colors = ['red', 'green', 'blue'];` instead of:
```
var name = 'colors';
var value = ['red', 'green', 'blue'];
```

2)  Is the following looping syntax still supported in marko-v3?
```
<ul>
    <li for="color in colors">${color}</li>
</ul>
```
When I run the above in my v3 test the generated output looks like this (and does not loop):
```
var colors = ['red', 'green', 'blue'];

    out.w("<ul><li>" +
      escapeXml(color) +
      "</li></ul>");
```
But when I use this syntax, it does work properly:
```
<div for(color in colors)>
    ${color}
</div>
```

PR #186, opened 2015-12-21
Wrapping var tag body in anonymous function

BryceEWatson:
For issue https://github.com/marko-js/marko/issues/169
Checks for Var tag body elements, wraps entire output in anonymous function if detected. Var tags with no body elements keep the previous behavior.
Test updated as well.


patrick-steele-idem:
Hey @BryceEWatson, what you provided should work, but I think it is best to avoid writing the code for the self-invoking function in `Var.js`. In this case, you are not adjusting the indentation correctly so the output code will be poorly formatted.

Instead, I would like to see the following:

```javascript
return builder.selfInvokingFunction([ this ]);
```

I realize that would result in an infinite loop unless you set a flag to make the node as already being wrapped. What are your thoughts of updating your code to use `builder.selfInvokingFunction([ this ])` and also setting a flag to avoid the infinite wrapping?

BryceEWatson:
Using `builder.selfInvokingFunction([ this ])` does make more sense. I've updated the PR.

patrick-steele-idem:
Thanks, Bryce! Changes look great

Issue #171, opened 2015-12-08
Marko v3: assign tag


patrick-steele-idem:
# New syntax

```xml
<assign foo=123 bar='hello'/>
```

Generated JavaScript code:

```javascript
foo=123;

bar='hello';
```

# Old syntax

```xml
<assign var="foo" value="123"/>
<assign var="bar" value="'hello'"/>
```

Issue #178, opened 2015-12-08
Marko v3: include tag


patrick-steele-idem:
NOTE: New syntax not finalized.

# New Syntax

```xml
<include("./include-target.marko") name='Frank' count=20/>
<include("./include-target.marko", {name: 'Frank', count: 20})/>
<include("./include-nested-content.marko") name="Frank" count=20>
    Have a 
    <b>
        wonderful
    </b>
    day!
</include>
```

NOTE: When body content is provided, a special `data.renderBody = function(out) { ... }` function should be added to the template data for the target template. 

# Old Syntax

```xml
<include template="./include-target.marko" name="${'Frank'}" count="20"/>
<include template="./include-target.marko" name="Frank" count="${20}"/>
<include template="./include-target.marko" template-data="{name: 'Frank', count: 20}"/>
<include template="./include-nested-content.marko" name="Frank" count="${20}">
    Have a 
    <b>
        wonderful
    </b>
    day!
</include>
```

BryceEWatson:
What do you think about keeping the template-data attribute in the new syntax? I like the new simplified syntax as well, I just wonder if template-data might be more intuitive for some?

patrick-steele-idem:
In general, I think we should avoid having two ways to do the same thing. I don't like that `template-data` would be a special attribute so I am in favor of only allowing template data to be the second argument inside the `<include(path, templateData)>` tag.

@philidem What do you think?

BryceEWatson:
@patrick-steele-idem, are there any examples of what the test file include-nested-content.marko would look like?

patrick-steele-idem:
Not yet @BryceEWatson, but I can add some code. Nested content should always be passed in as a `data.renderBody(out)` method.

patrick-steele-idem:
@BryceEWatson I went ahead and implemented support for `<include>` since I needed to move on to migration testing. Please review the code (commit: https://github.com/marko-js/marko/commit/a3d76a367842fd2171f481c1b8d4c72f12ac5a31) and let me know if you see any issues. Thanks for your help earlier.

PR #190, opened 2015-12-23
Adding assign tag

BryceEWatson:
For issue https://github.com/marko-js/marko/issues/171
Adds assign tag compiler, taglib, and test files.
The compiler uses `builder.assignment()` to write one or more assignments.


patrick-steele-idem:
Hey Bryce, thanks for the PR. What you submitted will work, but let's avoid introducing a new `Assign` AST node since it will confuse things (there is already an `Assignment` AST node). Instead, let's use the `codeGenerator(elNode, generator)` API for the custom tag to generate an array of assignment expressions. Something like the following should work:

```javascript
module.exports = function codeGenerator(elNode, generator) {
    var builder = generator.builder;

    return elNode.attributes.map((attr) => {
        return builder.assignment(attr.name, attr.value);
    });
};
```

As an example, the `<for>` tag uses the code generator API: https://github.com/marko-js/marko/blob/htmljs-parser/taglibs/core/for-tag.js

Instead of using `"node-factory"` in the `taglibs/core/marko-taglib.json` you should use do the following:

```json
"code-generator": "./assign-tag"
```

Also note, I submitted a bunch of changes to the `htmljs-parser` branch. Please make sure you pull in all of the latest changes! Thanks again!

patrick-steele-idem:
Hey @BryceEWatson, just wanted to check in and see if you had some time to try implementing the `<assign>` tag with the `code-generator` API. Please let me know if you would like more details. Thanks.

BryceEWatson:
Hi @patrick-steele-idem, I notice that Assignment does not add semi-colons or new lines. Assuming I'll just add those in the tag file instead? Or would this be something to add into the Assignment builder?

patrick-steele-idem:
The AST Nodes do not need to write out the semicolons. That gets handled by the `generateStatements()` method: 
https://github.com/marko-js/marko/blob/2a34d940e47e7617361ea7d7893927a46580d522/compiler/CodeGenerator.js#L289-L291

If you just return an array of assignment nodes then everything should work as expected, but please verify.

BryceEWatson:
Thanks @patrick-steele-idem, I've included your suggestion & squashed the commit. Let me know if there's any other improvements you see.

patrick-steele-idem:
Looks great. Glad everything works with less code. Thanks!

Let me know if you are interested in taking on the `<include>` tag next: https://github.com/marko-js/marko/issues/178

PR #223, opened 2016-02-11
Marko v3 - htmljs-parser + new compiler API

patrick-steele-idem:


yomed:
Awesome! :tada:

patrick-steele-idem:
Thanks @yomed. We are very close to a alpha release and I'm excited for others to start using the new Marko! Still a few more tasks left before the first alpha release: https://github.com/marko-js/marko/issues?q=is%3Aopen+is%3Aissue+milestone%3A3.0-alpha

There were a lot of commits across marko and [htmljs-parser](https://github.com/philidem/htmljs-parser), but we now have a much more extensive test suite (well over 500 individual tests and counting for marko, marko-widgets and htmljs-parser) so I feel free confident that things will work as expected.

We hope to have an alpha release early next week and the migration tool will come a little later.

PR #246, opened 2016-03-11
better error reporting during compilation of template

crsandeep:
Changes includes

- Better error reporting Ex: Unexpected identifier: (['red', 'blue', 'green'] foo)
- 'If statement' was not parsing the javascript expression, it is fixed now
- Added more negative tests
- Fixed the existing tests which broke due to changes with this PR

### DIFF
--- a/compiler/util/parseJavaScript.js
+++ b/compiler/util/parseJavaScript.js
@@ -1,6 +1,5 @@
 'use strict';
 var ok = require('assert').ok;
-var charProps = require('char-props');
 
 const esprima = require('esprima');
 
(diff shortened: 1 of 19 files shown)


compiler/util/parseJavaScript.js
```diff
@@ -219,19 +218,15 @@ function parseExpression(src, builder, isExpression) {
...
+        var errorMessage = '\n' + e.description;
         if (errorIndex != null && errorIndex >= 0) {
-            var srcCharProps = charProps(src);
             if (isExpression) {
                 errorIndex--; // Account for extra paren added to start
             }
-
-            let line = srcCharProps.lineAt(errorIndex)+1;
-            let column = srcCharProps.columnAt(errorIndex)+1;
-            errorMessage += ' [Line ' + line + ', Col: ' + column + ']';
```
yomed: Hi @crsandeep, any particular reason to remove line/col numbers here? They're probably still useful, even with the additional error message details.

patrick-steele-idem: Hey @yomed, I'll answer since I am reviewing the PR now. The problem is that the line and column number are for the expression only and don't match up with the line and column number within the template. In addition, there is no way to accurately convert the line and column number to a line and column number for the template since the expression gets modified from the original expression in the template to handle expanding placeholders.

Sandeep's solution is to simply show a marker at the column where the error was found on the line with the error. For example:

```
Unexpected identifier: (true sdsds)
                             ^
```

yomed: Ah gotcha, thanks for the details.

PR #250, opened 2016-03-18
Make renderSync behave if no context was supplied

jsumners:
This pull request fixes a case wherein if `renderSync` receives no context then it will still render the template. Currently, if you don't supply at least an empty object (`{}`) then `renderSync` bombs with an error like:

```
Debug: internal, implementation, error 
    TypeError: Uncaught error: Cannot read property '$global' of undefined
    at Object.Template.renderSync (/some/project/node_modules/marko/runtime/marko-runtime.js:109:17)
    ....
```


patrick-steele-idem:
Thanks for the improvement! I'll publish in just a bit

patrick-steele-idem:
New version published: `marko@3.0.3`

Thanks again.

jsumners:
Excellent. Thank you.

PR #261, opened 2016-04-05
Ability to exclude particular node_module from marko-taglib.json lookup

oxala:
* Added ability to exclude specific node_module from marko-taglib.json lookup;
* Added tests for excluding specific directory and specific node_modules from marko-taglib.json lookup;


patrick-steele-idem:
Thanks for the improvements! New version published: `marko@2.10.0`

PR #264, opened 2016-04-13
Replaced partial .MD URLs with absolute URLs

rbhogi:


patrick-steele-idem:
We need to use relative URLs so that docs will be correct on different branches. Thanks.

PR #278, opened 2016-05-01
Added docs folder and .jshint files to npmignore

crsandeep:
Adding some files to npmignore which are not required to be published to npm registry

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,1 +1,5 @@
 /test
+docs
+.jshintignore
+.jshintrc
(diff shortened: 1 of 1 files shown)


patrick-steele-idem:
Thanks for the PR but for convenience, we need to keep the docs folder as part of the npm package since the app that generates the docs for marko.json pulls in the `marko` npm package directly from npm to get the defaults. You can, of course, npm link in `marko`, but that adds a little friction.

PR #282, opened 2016-05-05
v3 attribute conditional syntax in language guide

briceburg:
the if="conditional" attribute syntax is deprecated in v3


patrick-steele-idem:
Good catch. Thank you for the fix!

PR #287, opened 2016-05-05
Add res.marko method for express apps

mlrawlings:
Adds a submodule (`marko/express`) that provides an `injectInto` method that will add a `marko` method to the express response prototype.  We get access to express locals, and it automatically sets the content type of the request.


patrick-steele-idem:
Looks great! What are your thoughts on renaming `injectInto(express)` to `patchExpress(express)`?

mlrawlings:
This latest update removes the need to pass `express` to the submodule.  It is required from the parent module's context.  Test coverage has been increased as well.

patrick-steele-idem:
Awesome. Thank you

PR #291, opened 2016-05-09
Emit async frament name on asyncFragmentFinish event

kprakasam:


patrick-steele-idem:
New version published: `marko@3.3.0`

Thanks!

PR #299, opened 2016-05-27
Add node context to errors thrown from a code-generator function

mlrawlings:
Ideally, all our errors in generating code would be developer friendly and added through `context.addError`, however, not every case is covered currently.  

This pull request catches errors from code generators, adds context about the node (including line number in the template when available), and rethrows the error while maintaining the original stack trace.

In the ideal case it looks something like this:
```
Error: Something Happened! -- Generating code for <custom-tag> tag (test/autotests/render/error-thrown-in-generator/template.marko:2:4)
```

Even in cases where we don't have the line number, you at least have a better idea of what might be causing the error.  The following is what the error @seangates encountered earlier would have looked like:
```
Error: 
Unexpected token ): (+)
                      ^ -- Generating code for <assign> tag
```
*Note: I'm actually not quite sure why the node doesn't have the line number attached in this case - I may look into this at a later point in time.*

For non-html nodes, the type of node is listed (and will list line number if available):
```
Error: some error message -- Generating code for ForEach node
```

### DIFF
--- a/compiler/CodeGenerator.js
+++ b/compiler/CodeGenerator.js
@@ -4,6 +4,7 @@ const isArray = Array.isArray;
 const Node = require('./ast/Node');
 const Literal = require('./ast/Literal');
 const Identifier = require('./ast/Identifier');
+const HtmlElement = require('./ast/HtmlElement');
 const ok = require('assert').ok;
 const Container = require('./ast/Container');
 const util = require('util');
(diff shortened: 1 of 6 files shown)


seangates:
@mlrawlings Wonderful! Thanks!

patrick-steele-idem:
Thanks for the PR @mlrawlings!

New version published with improvement: `marko@3.4.2`

PR #300, opened 2016-06-03
<macro> must have a name

mlrawlings:
`<macro>` now expects its first attribute to be its name (regardless of whether it has arguments or not), if there are no attributes or the first attribute has a value (ex. `test=123`), then it throws an error.

✅ good:
```xml
<macro greeting>
    Hello!
</macro>

<macro greeting()>
    Hello!
</macro>

<macro greeting(name)>
    Hello ${name}!
</macro>
```

:x: bad:
```xml
<macro async=true>
    Uh-oh!
</macro>

<macro test=123 greeting(name)>
    Uh-oh ${name}!
</macro>

<macro>
    Uh-oh!
</macro>
```

### DIFF
--- a/taglibs/core/macro-tag.js
+++ b/taglibs/core/macro-tag.js
@@ -1,13 +1,10 @@
 module.exports = function codeGenerator(elNode, codegen) {
 
     var attributes = elNode.attributes;
-    if (!attributes.length) {
-        return;
-    }
-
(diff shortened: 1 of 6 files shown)


patrick-steele-idem:
Thanks! New version published with improvement and CHANGELOG.md updated.

PR #301, opened 2016-06-03
Coverage

mlrawlings:
Adds coverage reporting through istanbul and coveralls.  We'll need to set up marko on [coveralls](https://coveralls.io/), but it's a pretty painless process.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,3 +5,4 @@ node_js:
 - "6"
 language: node_js
 script: "npm test"
+after_success: "npm run coveralls"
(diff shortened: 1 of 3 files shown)


yomed:
Nice PR @mlrawlings! Did you consider using `--include-all-sources` in the istanbul command for  coverage against the whole codebase?

patrick-steele-idem:
Awesome! Thank you for taking the time to do that. I have finished setting up coveralls and now we have a working badge. Thanks again!

> Did you consider using `--include-all-sources` in the istanbul command for coverage against the whole codebase?

@yomed: Would you be interested in trying that out to see if it makes the coverage more accurate?

PR #308, opened 2016-06-14
fixes #307

### LINKED ISSUE #307: Marko concise syntax, with multiple class names
Marko concise syntax seems to only allow two class names in the format `div.foo.bar`. For example, `div.foo.bar.another` does not work and outputs `<div class="foo bar"></div>`

mlrawlings:
Only set `prevClassName` when the `className` has been pushed to `finalClassNames`, otherwise future `className` values will be appended to a class that doesn't get included in the final output.  This effectively limits the number of classes to 2.


patrick-steele-idem:
Thanks for the fix! New version published: `marko@3.4.9`

PR #315, opened 2016-06-23
Issue 314 - Remove hyphens from include properties

### LINKED ISSUE #314: Attributes being passed to the include tag do not have hyphens removed
For example:

```xml
<include("./include-target.marko") first-name='Frank' count=20/>
```

That is currently being compiled to the following:

```javascript
__include_target.render({
        "first-name": "Frank",
        count: 20
      }, out);
```

It was supposed to be compiled to the following:

```javascript
__include_target.render({
        firstName: "Frank",
        count: 20
      }, out);
```

As a result of this bug, the property needs to be read from the target template as `${data['first-name']}` instead of the simpler `${data.firstName}`.

patrick-steele-idem:
- Fixes #314 - Remove hyphens from include props
- Deprecated:
    - Properties passed in using the `<include>` tag should not be access using hyphens.

For example, given the following template:

```xml
<include("./include-target.marko") first-name='Jane'/>
```

The `first-name` data should be accessed using the `firstName` property:

```javascript
var firstName = input.firstName;
// NOT: var firstName = input['first-name'];
```


mlrawlings:
Looks good!

Issue #319, opened 2016-06-27
Hot reload fails if NODE_ENV is not 'development' or 'dev'


ianvonholt:
I tracked down a bug with hot-reloaded templates not re-compiling if NODE_ENV is set to values other than "development" or "dev". 

The error stems from: 
https://github.com/marko-js/marko/blob/master/runtime/loader.js#L102-L108
and
https://github.com/marko-js/marko/blob/master/compiler/config.js#L24

The assumeUpToDate variable should take into consideration that the files are being hot-reloaded before taking the node environment into consideration.

PR #320, opened 2016-06-27
Adding detection for process.env.MARKO_HOT_RELOAD with assumeUpToDate…

ianvonholt:
… configuration variable. If files are hot reloaded, we should never assume that they are up to date.

Fixes issue #319


patrick-steele-idem:
Thanks for the improvement! New version published with fix: `marko@3.6.2`

Please let me know if you see any other issues. Thanks, again.

PR #332, opened 2016-07-17
change value of literalFalse to false

bkuri:
HTH


mlrawlings:
@bkuri This one also fails due to the test being incorrect... oops!

PR #333, opened 2016-07-17
change value of literalUndefined to undefined

bkuri:
HTH


mlrawlings:
@bkuri Thanks! Since you've already got a branch for this, would you mind running the tests (`npm test`) and fixing the one that fails?  It's actually asserting that `literalUndefined === null`! 😆

PR #338, opened 2016-07-18
Add autodiscover of components/ directory. fixes #329

### LINKED ISSUE #329: Config Free Taglib Discovery
@aniruddh-a and I were discussing how custom tags are discovered, and he wondered why it couldn't just be done automatically.  This might be a good idea to explore.

Essentially, it would function kinda like `node_modules` except based on the `components` directory (which has become a standard way of doing things).  The taglib lookup could search first in the current directory for a `components` directory and continue walking up to the project root to discover tags.

This actually mirrors the way that `marko.json` files are discovered and could be done in the same step.

Thoughts?

/cc @marko-js/maintainers

mlrawlings:
I'm testing this using a render test.  Is that fine?


patrick-steele-idem:
I think a render test is great because it easy to write tests based on the output HTML. Alternatively, you could add a taglib lookup test if you felt the need to inspect some of the internal data structures similar to the following: https://github.com/marko-js/marko/blob/master/test/autotests/taglib-lookup/custom-tag/test.js

patrick-steele-idem:
Amended commit has been merged: https://github.com/marko-js/marko/commit/80999593551b08d5a838daed462eb7561a735ec5

New version published with improvement: `marko@3.8.0`

Closing PR.

PR #371, opened 2016-10-01
add docs for while loop

yomed:
`while` was introduced in v3, but was missing from the docs. I copied this from the "what's new" page, and then modified it to the same format as the `for` loop documentation.


patrick-steele-idem:
Looks great. Thanks!

PR #411, opened 2016-11-09
Fixes #407 - Allow multiple extensions when installing the Node.js re…

### LINKED ISSUE #407: Allow multiple extensions when installing the Node.js require hook
Currently, only a single extension can be provided. We should add support for a new `extensions` option:

```javascript
require('marko/node-require').install({ extensions: ['.marko', '.marko.xml', '.html'] });
```

austinkelleher:
`extension` and `extensions` properties are both now supported and consolidated. A few things to discuss:

- The consolidation function takes a `requireObj`, which is the Node.js `require` in `node-require.js`. I did this so it was easy to mock in a unit test without having to pollute the actual `require` object.
- I put `consolidateExtensions.js` in a new folder `util/`. I am open to other suggestions as to where to put this module. I do like the idea of having a `util/` folder at the top-level though.

### DIFF
--- a/node-require.js
+++ b/node-require.js
@@ -13,11 +13,13 @@
 * See the License for the specific language governing permissions and
 * limitations under the License.
 */
+'use strict';
 
-var path = require('path');
-var resolveFrom = require('resolve-from');
-var fs = require('fs');
-var fsReadOptions = { encoding: 'utf8' };
+const path = require('path');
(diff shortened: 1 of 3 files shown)


patrick-steele-idem:
Changes already manually merged. Closing PR. Thanks for working on this!

PR #423, opened 2016-11-11
Fixes #251 - Output of render should be Promise compatible.

### LINKED ISSUE #251: Output of render should be Promise-compatible
Currently rendering a template returns an `AsyncWriter`, it would be handy though if when a callback is omitted that templates render function returned a promise.

``` js
var template = require("template.marko");

template.render({ hello: "world" })
    .then(function (html) { ... })
    .catch(function (err) { ... });
```

If you're using promises already it makes using marko a bit easier.

austinkelleher:


patrick-steele-idem:
If `out.error()` is called the `finish` event is still emitted. That would result in both `reject` and `resolve` being called. I'm not sure what the `Promise` spec says should happen when that occurs, but we should probably add code to make sure that `resolve` is not called when `reject` is called. Do you mind investigating that?

Hesulan:
The `Promises/A+` specification requires that a `Promise` instance must only ever handle the very first call to either `resolve` or `reject`, and ignore all further attempts at either. So as long as the correct one is called first it doesn't matter what happens after, the resolved/rejected `Promise` instance becomes immutable.

Hesulan:
That being said, do be aware that a fulfilled `Promise` exhibits *shallow* immutability; the resolved/rejected value is effectively a `const` in that you cannot *replace* it, but you can still modify the result's properties. And since calling `resolve` or `reject` doesn't cause the function to exit, it's entirely possible to accidentally mutate the result.

test/AsyncStream-test.js
```diff
@@ -287,6 +287,22 @@ describe('async-writer' , function() {
...
+            expect(err).to.be.an('error');
+            expect(out.getOutput()).to.equal('1');
+            done();
+        }).then((data) => {
+            throw new Error('Should not get here!');
+        });
+
+        out.write('1');
+        out.error(new Error('test'));
+        out.write('2');
```
patrick-steele-idem: Aren't you missing an `out.end()`?

PR #428, opened 2016-11-13
Added 'const' and 'let' tags

Hesulan:
## Description
I simply copied `var-tag.js` and its `marko.json` entry to create `<let>` and `<const>` tags in `taglibs/core`.

## Motivation and Context
I was surprised to find that `context.builder.vars()` already had an optional second parameter for variable type ('var', 'let', or 'const'), but only `<var>` had been implemented as a tag.

This pull request is meant to be a starting point, but it could also be applied as-is. The commit is purely additive and no existing functionality should be affected.

## TODO
* Add automated tests
* Add documentation
* (optional) Throw compiler error on identifier re-declaration or on `const` re-assignment

## Additional Notes
Providing a body to a `<var>` tag causes it to compile into an IIFE to restrict its scope. Because the `<let>` and `<const>` tags simply re-use the same code they currently do the same, but they could be made to use a block statement (`{ ... }`) instead.


Hesulan:
Hmm... it seems the auto-tests are currently failing, I assume that's because I didn't write any. Unfortunately I'm not sure when I'll next get the chance to work on this. If anyone would like to take over from here, feel free to do so.

patrick-steele-idem:
Thank you @Hesulan! For completeness, I think it is great that you worked on this. The tests are failing because we have some tests for the "taglib lookup" that will fail if you add new core tags. You simply need to update the following two files to include your new core tags (i.e. `const` and `let`):

- https://github.com/marko-js/marko/blob/master/test/autotests/taglib-lookup/forEachTag/expected.json
- https://github.com/marko-js/marko/blob/master/test/autotests/taglib-lookup/getTagsSorted/expected.json

(you just need to copy the text from `actual.json` over to `expected.json`)

Please do add tests when you get a chance. Copying and modifying the following test is sufficient for now: https://github.com/marko-js/marko/tree/master/test/autotests/render/var

austinkelleher:
I don't really like the idea that `var`, `let`, and `const` are treated the same by the compiler. I think if we introduce these new tags, they need to function differently. For example this is okay with `var`:

```html
    <var nameUpper=data.greetingName.toUpperCase() nameLower=data.greetingName.toLowerCase()>
    	Hello ${nameUpper}!
	    Hello ${nameLower}!
        <var nameUpper=data.greetingName.toLowerCase() nameLower=data.greetingName.toUpperCase()>
    		Hello ${nameUpper}!
	    	Hello ${nameLower}!
    	</var>
    </var>
```

But it should throw a compile time error using `let`:

```html
    <let nameUpper=data.greetingName.toUpperCase() nameLower=data.greetingName.toLowerCase()>
    	Hello ${nameUpper}!
	    Hello ${nameLower}!
        <let nameUpper=data.greetingName.toLowerCase() nameLower=data.greetingName.toUpperCase()>
    		Hello ${nameUpper}!
	    	Hello ${nameLower}!
    	</let>
    </let>
``` 

`const` should also be treated like like a `const` in JavaScript, so if we declare `<const>` and try to use the `<assign>` tag below it, the compiler would throw an error:

```html
<const nameUpper=data.greetingName.toUpperCase() nameLower=data.greetingName.toLowerCase()>
    	Hello ${nameUpper}!
	    Hello ${nameLower}!
        <assign nameUpper=data.name.toLowerCase()/>
    </const>
```

What do you think @patrick-steele-idem / @mlrawlings?

patrick-steele-idem:
@austinkelleher, `let` and `const` would compile to code that uses `let` and `const`. 

Also, just wanted to point out that `let` is beneficial in the following contrived situation:

```
<let foo='Frank'/>

<if(someCondition)>
    <let foo=123/>
    <!-- do something with foo -->
</if>
<else>
    <let foo=456/>
    <!-- do something with foo -->
</else>

<div>
  Hello ${foo}! <!-- Outputs "Hello Frank!" -->
</div>
```

Within the body of custom tags, `var` happens to be scoped correctly since marko automatically wraps the body of custom tags in a `renderBody(out)` function.

Hesulan:
I agree that the compiler should throw an error on re-initialization or const re-assignment, that's point #3 in the TODO section of the pull request. It's getting compiled into actual `let` and `const` calls so it will eventually do that during render anyway, but it would be nice if it could do that at compile-time. But that's something that will need to be done deeper within the code, probably somewhere inside the `Vars` AST node; so far these commits have only touched the taglib code.

I'm not sure why I didn't think of this before, but it will also be necessary for the compiler to handle `let` and `const` differently in pre-ES6 environments, where the keywords either need to be transpiled, or require strict-mode (as two of the four travis-ci tests apparently do). In any case, it seems this isn't quite as simple as it appeared.

Assuming ES6 or higher, anyone can grab these commits to play with locally, it's worked perfectly for me so far. I might also get around to making this a separate taglib in the next day or two, at least until someone with more time on their hands (possibly my future self? I can only hope...) figures out how to make it not break things.

patrick-steele-idem:
@mlrawlings Can you please review this PR and decide on next steps? Thanks.

Hesulan:
So I've been mulling this over, and here are the solutions I can think of:

1. Pull these tags into a separate module. This would be fairly simple, I would just need to make a few minor changes and publish to npm.

2. Provide built-in support for compiling to `var` instead. This would be overkill, and I'm not even sure where to begin.

3. Add `'use strict';` to the top of the compiled output. This would make it work correctly in Node 4 and 5, but would also change the behavior of the entire template in very subtle ways, albeit mostly for edge cases.

4. Expect end users to know that they can't use these tags natively in pre-ES6 environments. We would also need to modify the integration tests to skip certain ones on Node 4 and 5.

5. Pull these tags into a new built-in taglib specifically made for support-breaking features (something like `es6-taglib`) which can then be enabled/disabled by a compiler flag (like `require('marko/compiler').defaultOptions.es6_tags = true`). This would be especially helpful in the case of client-side rendering, where templates might be rendered in a much lower environment version than the one they were compiled in, and developers shouldn't have to keep track of which tags support which versions. Any future tags which generate ES6+ code would also be placed here.

I'm sure there other options I haven't considered as well. Either way, I'd like to hear your thoughts!

mlrawlings:
I'm torn on what to do here.  These are language level features, so I don't think they belong in an external package.  If we're going to support this, I don't want to handle transpiling.  I'd either be in favor of letting it pass through unchanged (node 6+), or if we want to support 4/5 prepend `'use strict';` to the template.

On the other hand, I'm not sure we really need them.  The only real benefits are they introduce a temporal dead zone and `const` can't be redefined.

And taking Patricks example from above, we already have a way to create a TDZ of sorts:
```html
<var foo='Frank'/>

<if(data.someCondition)>
    <var foo=123>
        <!-- do something with foo -->
    </var>
</if>
<else>
    <var foo=456>
        <!-- do something with foo -->
    </var>
</else>

<div>
  Hello ${data.foo}! <!-- Outputs "Hello Frank!" -->
</div>
```

The `es6-taglib` taglib idea is interesting, but I don't think it's necessary.  We will always be able to write something that is es5 compatible for core taglibs.  It's only these places where we're trying to emulate the javascript language.

Also, userland tags might require es6, and we don't have control over that, but also I'm not sure we want to introduce something at the core that when used in userland modules forces the application using them to transpile or drop support for older platforms.

Hesulan:
For me, the point of `const` and `let` is that they're clean, effective, *native* solutions to specific problems. A separate taglib or module would provide a safe, "official" place for tags like these which are built around ES6+ features and concepts to be implemented without the risk of accidentally breaking compatibility; anyone who chooses to `require()` it knows what they're getting into. That being said, the addition of another taglib or module to maintain isn't something to propose lightly.

In any case, this certainly isn't urgent. Perhaps we could simply put this on the back-burner while we experiment with ideas for other ES6+ tags, and see how much interest there is from the community? In the meantime, I could easily port these to a separate module and publish it for everyone to play with.

mlrawlings:
@Hesulan I think that makes sense.  `marko-layout` and `marko-async` started out as separate packages that eventually were merged into core.  The same could happen here.  I'm going to go ahead and close this issue, but do publish that package so we can play with it.  And maybe we can reopen this in the future.

Hesulan:
For anyone interested, I finally got around to publishing these as a standalone package: 
```
npm install --save marko-taglib-let-const
```
(Note: The `--save` part is necessary because marko only searches for taglibs listed as dependencies in your `package.json`.)

PR #433, opened 2016-11-15
Fixes #419 - Deprecate w-preserve/w-preserve-if/w-preserve-body/w-pre…

### LINKED ISSUE #419: Deprecate w-preserve/w-preserve-if/w-preserve-body/w-preserve-body-if
`w-preserve` → `no-update`

**Old:**
```html
<div w-preserve>
    ...
</div>
```

**New:**
```html
<div no-update>
    ...
</div>
```

austinkelleher:


austinkelleher:
I'm not sure what's up with Coveralls. It's running twice.

PR #435, opened 2016-11-16
Stream without a context provided (fix exception to undefined variable)

jsumners:
## Description

```js
var template = require('some.marko')
template.stream() // exception on line 234 of runtime/marko-runtime.js
```

## Motivation and Context

Fixes a runtime exception when streaming without providing a data context.

I would provide a test for this, but I can't find where it would go or any tests that resemble the case to emulate.


austinkelleher:
I think a test for this should live in `test/autotests/api`. An example is: `test/autotests/api/require-render-to-stream`.

jsumners:
@austinkelleher thanks. I added the test.

austinkelleher:
Also, the API has changed in v4, and I don't think this is an issue anymore, but I think this change should be back ported to v3.

jsumners:
@austinkelleher I wrote this against the v3 branch. It is a bug that is impacting my current usage of Marko.

austinkelleher:
LGTM /cc @mlrawlings

austinkelleher:
Actually I have a small suggestion after reviewing the code in v4. What do you think about changing the implementation to be more uniform with the v4 implementation?

https://github.com/marko-js/marko/blob/master/runtime/stream.js#L40

mlrawlings:
I like it.  Thanks @jsumners and @austinkelleher!

mlrawlings:
@jsumners Just published `marko@3.12.1`

jsumners:
@mlrawlings thank you.

Issue #442, opened 2016-11-21
The `.then` and `.catch` methods in `runtime/OutMixins.js` are broken


Hesulan:
I finally got around to looking at the newly added `.then` method from #423 , and it turns out that it's nowhere near spec-compliant: attempting to call `.catch` causes an error to be thrown ("TypeError: fn is not a function", referring to the `undefined` passed as the first argument to `.then(fn, fnErr)`), the handlers are called synchronously, ~~and `.then` itself ceases to function properly once the stream is finished (subsequent calls return a Promise resolved with `undefined`).~~

~~There are only 3 possible solutions I can think of:~~
~~1. Let `.then` be non-compliant. It will only work if called *before* the stream is finished, among other things.~~
~~2. Attach a Promise instance to the stream regardless of whether `.then` is ever called or not. The renderer would then always require a native or polyfilled `Promise` implementation, even if `.then` is never actually used.~~
~~3. Write a partial `Promise` implementation directly into the renderer. I could do this if needed, but it would probably be better to simply require a native or polyfilled implementation.~~

EDIT: It seems I was mistaken about it ceasing to function after 'finish' is called, though it doesn't quite appear to function correctly in the first place.

EDIT 2: I'm actually getting all sorts of weird problems from this code and I'm not entirely sure what's causing most of them. In case someone else wants to take a stab at it, here's a spec-compliant solution, but it caused the test suite to throw an out-of-memory error when I tried it:
```js
then: function(fn, fnErr) {
    var self = this;
    return new Promise(function(resolve, reject) {
        self.on('error', reject);
        self.on('finish', resolve);
    }).then(fn, fnErr);
}
```

Hesulan:
So I just now realized why that was throwing an error: The stream itself is being passed to 'finish', which is then passed to `resolve`, which attempts to call the stream's `.then` method, which leads to an infinite loop. It needs to instead pass the result string to `resolve`, but the test suite currently depends on the resolved value being the stream itself (which *definitely* violates the Promises/A+ spec, since the stream is a thenable).

In other words, the tests need to be rewritten. Then the above solution, with `self.on('finish', function(data){resolve(data.getOutput());});` instead of `self.on('finish', resolve);`, should work perfectly. (I'm *pretty* sure it's `data.getOutput()`, but double check that...)

I won't get the chance to do that tonight, or probably tomorrow either, so if anyone else would like to do the honors, be my guest. If not, I'll hopefully get to it this week.

[...]

Issue #439, opened 2016-11-17
Promises/A+ Compliance


Hesulan:
While digging around in the compiler just now I noticed that some of the code is *not* Promises/A+ compliant, specifically the `<await>` tag. These are mostly edge-cases, and will *probably* never break anything, but here's what I found:

* The promise resolution procedure must only access the `.then` property once. So instead of `if (typeof promise.then === 'function') { promise.then(...); }` it should use `var then = promise.then; if (typeof then === 'function') { then.call(promise, ...); }`

* If attempting to retrieve the `.then` property throws an error, then the error must be caught and the resulting promise must be rejected with the error as the reason. (It currently isn't being caught.)

* The `.catch(onRejected)` method is neither defined nor required by the specification, it's simply a convenient alias for `.then(null, onRejected)` which *most* promise implementations *choose* to provide. 

Those are just the things I happened to remember, there are probably a few other technicalities that I've forgotten about since I last read through the spec. I'll try to work on this when I get the chance but until then I thought I should post this somewhere. I'll also keep an eye out for other places in the code that might handle Promises incorrectly and add them here.

Promises/A+ Specification: https://promisesaplus.com/

Hesulan:
This also raises an interesting question: Should Marko actually be trying to handle Promise resolution directly, or would it be better to rely on a native/polyfilled `Promise` implementation? It looks like that's what #251 is currently doing.

If Marko is to handle "thenables" directly, then I would be in favor of adding a common helper utility for it - which would essentially be a built-in polyfill - so that each tag or feature which needs to handle Promise resolution won't end up re-implementing their own non-compliant handling code. I have enough experience writing spec-compliant thenables to do that, but I would much prefer letting a dedicated `Promise` implementation handle it.

Hesulan:
Quick update: Since #251 only applies when the user explicitly calls the `.then` method, it's considered a special case. The `<await>` tag, however, needs to be usable in pre-ES5 environments, which leaves either a built-in polyfill or just more compliant handling code as the only options.

However, I've also noticed one other issue: The `<await>` tag currently accepts either a value, a Promise, or a node-style function (last argument is an error-first callback). I'll continue working on this under the assumption that we want the `<await>` tag to continue to handle node-style functions, but be aware that this behavior defies the Promises/A+ specification.

patrick-steele-idem:
Here are some replies:

> The promise resolution procedure must only access the `.then` property once.

Why do you say that?

[...]

PR #443, opened 2016-11-22
updated render api. fixes #389.

### LINKED ISSUE #389: Consistent rendering API
**Everything returns an `out`**

Currently, `template.render()` returns `out` and `template.renderSync()` returns the html string.

We will change `renderSync` to also return `out`.  This will be a breaking change, but we will  implement `out.toString()` and `out.toJSON()` so in many cases the out can be used as a string.

`typeof out === 'string'` returns `false` and you can't use the `String.prototype` methods on `out`, but you will be able to pass the contents to `fs.writeFile`, `res.send`, `JSON.stringify`, etc. and it still works as if it were a string (in many cases).

**DOM helpers**
Marko Widgets (which has now been merged into this repo) had implemented it's own renderer using which returned a `RenderResult`: a lightweight wrapper around `out` than implemented helper methods like `.appendTo()` and others.  

Widgets will no longer have a special render.  They will return `out` just like templates, but we have added these DOM helpers to all instances of `out`, so you can use these for templates as well as widgets.

**Sync Errors**
For when rendering using `renderSync`, we already throw an Error if anything tries to `beginAsync`.  However we don't do the same on the other side: if you call `template.render({}).appendTo()` and the template has not finished rendering yet, the call to `appendTo()` should throw.

**Events, Promises, Streams, & Callbacks**
Async templates will be handled by `template.render({}).on(‘finish’)`, `template.render({}).then()`, `template.render({}, stream)`, or `template.render({}, callback)` all of which are passed the `out` instance.

mlrawlings:
## Description

Making the render api consistent.

### Sync Render
Rendering synchronously now returns a RenderResult:
```js
var result = template.renderSync({});
result.appendTo(document.body);
result.getWidgets();
// etc.
```

**Differences:**
In v3, we return a string.  In order to make this a more painless switch, I will introduce a `renderToString` method in v3 and prompt users to switch to using that if they need that string output.

### Async Render

Calling `.render()` still returns an out:
```js
var out = template.render({});
out.on('finished', function(out) {
    
});
```

**Differences:**
In v3, there is no value passed to the `finished` event.  No breaking change here.

### Callback 

```js
template.render({}, function(err, result, out) {

});
```

**Differences:**
In v3, the callback is passed the `html` string instead of a RenderResult.  The third parameter, `out` is only passed for legacy reasons.  You can also access it from `result.out`.

### Promise

See also #439, #442.

```js
template.render({}).then(result => result.appendTo(document.body));
```

**Differences:**
In v3, then was not available.  No breaking change here.

### Widgets

`defineRenderer` and `makeRenderable`/`renderable` now include both `render` and `renderSync` that function similarly to above.  I plan to add `renderSync` to `marko-widgets@6` and warn users that are using the api like:

```js
widget.render({}).appendTo(document.body);
```

to switch to 

```js
widget.renderSync({}).appendTo(document.body);
```
or
```js
widget.render({}, function(err, result) {
    result.appendTo(document.body);
});
```

## Checklist:

@patrick-steele-idem I'd like some eyes on this and the plan described above before merging.
@Hesulan How does the Promise implementation look?

### DIFF
--- a/runtime/RenderResult.js
+++ b/runtime/RenderResult.js
@@ -1,14 +1,14 @@
 var events = require('./events');
 var dom = require('./dom');
 
-function checkAddedToDOM(asyncStream, method) {
-    if (!asyncStream.data._added) {
+function checkAddedToDOM(result, method) {
+    if (!result.out.data._added) {
         throw new Error('Cannot call ' + method + '() until after HTML fragment is added to DOM.');
     }
 }
 
-function getWidgetDefs(asyncStream) {
-    var widgetDefs = asyncStream.data.widgets;
+function getWidgetDefs(result) {
+    var widgetDefs = result.out.data.widgets;
 
     if (!widgetDefs || widgetDefs.length === 0) {
         throw new Error('No widget rendered');
@@ -16,11 +16,15 @@ function getWidgetDefs(asyncStream) {
     return widgetDefs;
 }
 
-module.exports = {
+var RenderResult = module.exports = function RenderResult(out) {
+    this.out = out;
+};
+
+RenderResult.prototype = {
     getWidget: function() {
         checkAddedToDOM(this, 'getWidget');
 
-        var rerenderWidget = this.global.__rerenderWidget;
+        var rerenderWidget = this.out.global.__rerenderWidget;
         if (rerenderWidget) {
             return rerenderWidget;
         }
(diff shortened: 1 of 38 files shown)


Hesulan:
Doesn't this cause "finish" to return a new RenderResult, making the "self.on('finish', function(){ resolve(new RenderResult(out)); })" redundant?

Either way, looks great to me! I'll look a bit closer tomorrow just to be sure, but I believe this should work perfectly, at least concerning Promise compliance.

Hesulan:
(In other words, wouldn't "self.on('finish', resolve);" do the same thing? Or I might just have mis-read something.)

Hesulan:
Nevermind, was half asleep when I typed that, I did mis-read something. Just skimmed over the code again, it all looks good to me!

PR #444, opened 2016-11-22
The `<await>` tag should attempt `.then(undefined, ...)` when resolving a thenable if neither `.catch` nor `.fail` is a function

Hesulan:
As discussed in #439, the `Promises/A+` specification explicitly defines `.then` as accepting at least two arguments: `onResolved` and `onRejected`. The `.catch` method is defined by ECMAScript for *native* implementations, but third-party libraries are under no obligation to include it (though *most* do, for convenience and compatibility).

In the interest of caution, this only calls `.then(undefined, onRejected)` if neither `.catch` nor `.fail` is a function. However, in my opinion, it would be completely safe to remove those two calls altogether, as any implementation which is even *remotely* compliant will handle this as expected.

### DIFF
--- a/taglibs/async/await-tag.js
+++ b/taglibs/async/await-tag.js
@@ -24,7 +24,7 @@ function promiseToCallback(promise, callback, thisObj) {
                 callback(err);
             });
         } else {
-            finalPromise = finalPromise.then(void 0, function(err) {
+            finalPromise = finalPromise.then(null, function(err) {
                 callback(err);
             });
         }
--- a/taglibs/async/await-tag.js
+++ b/taglibs/async/await-tag.js
@@ -23,6 +23,10 @@ function promiseToCallback(promise, callback, thisObj) {
             finalPromise = finalPromise.fail(function(err) {
                 callback(err);
             });
+        } else {
+            finalPromise = finalPromise.then(void 0, function(err) {
+                callback(err);
+            });
         }
 
         if (finalPromise.done) {


taglibs/async/await-tag.js
```diff
@@ -23,6 +23,10 @@ function promiseToCallback(promise, callback, thisObj) {
             finalPromise = finalPromise.fail(function(err) {
                 callback(err);
             });
+        } else {
+            finalPromise = finalPromise.then(void 0, function(err) {
```
patrick-steele-idem: Do you mind switching from `void 0` to `undefined`? If `undefined` !== `void 0` then we would be in a huge mess

Hesulan: The point of using `void 0` is that it *always* produces a literal `undefined` (ever since the very first version of javascript), while the `undefined` variable can be redefined.

Hesulan: I can certainly change this if desired, all ES5+ environments will handle both identically (in fact I vaguely remember reading somewhere that `undefined` gets compiled into `void 0` for performance reasons).

mlrawlings: As of IE9, the global `undefined` is immutable (you can still create a local variable named `undefined`).  

`void 0 === undefined` is safer as `void` is a reserved word and cannot be overridden.  I do think `undefined`/`null` reads better though.

Interestingly, the native `Promise` implementation doesn't care what you pass.  If it's a function it'll call it, otherwise not.  `promise.then('foo', console.log)` works perfectly fine.

Hesulan: As defined by the Promises/A+ specification (either explicitly or implicitly, I can't remember which), `onResolved` and `onRejected` are simply ignored if they're not a function. The only concern would be if a function was assigned to `window.undefined`, which wouldn't be possible in IE9+ or other modern browsers.

I do agree that `undefined`/`null` is easier to read. I would be in favor of `null`, as long as it wouldn't break any of the popular libraries (which it shouldn't).

patrick-steele-idem:
Thanks! 👍

PR #447, opened 2016-11-23
Fixes #446 - Normalize the `expected` and `actual` strings during autotests

### LINKED ISSUE #446: Proper ending newlines (as required by POSIX) break the test suite
### Description
The autotests work correctly when the `expected.*` file is created and saved on Windows. However, Linux and OS X comply with POSIX, which requires a proper ending newline (`\n` immediately followed by the `EOF` symbol). The produced `actual.*` file doesn't have one, which causes the autotests to fail.

### Possible Fix
The autotest runner could simply ignore the presence or absence of an ending newline. Alternatively, the rendered output could be made POSIX-compliant (by adding a proper EOF newline), but Windows displays it as a blank line and allows it to be removed which could lead to confusion.

### Steps to Reproduce
1. Open an `expected.*` file from the autotest directory in any text editor
2. Do one of the following:
  a. (Linux or OS X) Save the file with no changes
  b. (Windows) Insert a blank newline at the end of the file and save the changes
3. Run the test suite

### Example
Saving `test/autotests/render/assign-tag/actual.html` on Linux with no changes and running `npm test` produces:
```
AssertionError: 'HELLO 1230' deepEqual 'HELLO 1230\n'
```

### What does Git have to say about all of this?
Git expects there to be an ending newline (because there really *should* be), and has to internally add `"\ No newline at end of file"` as a workaround, shown by github as a red symbol on the last line - go look at the commit diff for almost any file written on Windows, you'll see what I mean.

### Workarounds
The only workaround I can find is to delete the last line and use `echo -n ... >>` to reinsert it without a trailing newline.

Hesulan:
The autotest `compareHelper` now normalizes the ending newline of both the `expected` and `actual` strings to avoid test failures when `expected.*` is saved on a POSIX-compliant (non-Windows) OS.

### DIFF
--- a/test/autotest.js
+++ b/test/autotest.js
@@ -39,11 +39,9 @@ function compareHelper(dir, actual, prefix, suffix) {
         fs.writeFileSync(expectedPath, expectedString, {encoding: 'utf8'});
     }
 
-    if (isObject) {
-        actual = JSON.parse(actualString);
-    }
+    actual = isObject ? JSON.parse(actualString) : actualString.replace(/\r?\n$/, '');
 
-    var expected = isObject ? JSON.parse(expectedString) : expectedString;
+    var expected = isObject ? JSON.parse(expectedString) : expectedString.replace(/\r?\n$/, '');
     assert.deepEqual(actual, expected);
 }
 


patrick-steele-idem:
Thanks! 👍

PR #448, opened 2016-11-27
add <import> tag Fixes #404

### LINKED ISSUE #404: Introduce the `<import>` tag for ES6 imports
```xml
<import helpers from "./helpers"/>
- Total: ${helpers.formatCurrency(data.total))

// Concise syntax:
import helpers from "./helpers"
- Total: ${helpers.formatCurrency(data.total))
```

The `<import>` should follow the same rules as an ES6 import and the imports should be moved to the top of the compiled template.

Open questions:
- Should marko transpile the ES6 import code to CommonJS require?

(As a side note, I'm going to continue to call it ES6 instead of ES2015.)

scttdavs:
## Description

adds import tag, works like: 
```html
<import mod, { b as bar } from "./bar" />
```

## Motivation and Context
Fixes https://github.com/marko-js/marko/issues/404

## Checklist:


patrick-steele-idem:
Thanks for working on this @scttdavs!  I did a fairly quick code review and one issue is I see is that we want to avoid parenthesis when using the `<import>` tag so that the user can write out the import in the ES6 way. Currently, it appears you are parsing the import statement from within the argument on the `<import>` tag:

```xml
<import({bar as b} from './bar') />
```

I would like to only support the following:

```xml
<import {bar as b} from './bar' />
```

The only code change that should be required is that you will need to loop over the attributes and join the attribute names by a space (e.g., `attributeNames.join(' ')`) to build the import string that is then parsed. For example, here is the AST after parsing `<import {bar as b} from './bar' />`:

```json
{
    "type": "TemplateRoot",
    "body": [
        {
            "type": "HtmlElement",
            "tagName": "import",
            "attributes": [
                {
                    "type": "HtmlAttribute",
                    "name": "{bar as b}"
                },
                {
                    "type": "HtmlAttribute",
                    "name": "from"
                },
                {
                    "type": "HtmlAttribute",
                    "name": "'./bar'"
                }
            ],
            "body": []
        }
    ]
}
``` 

Do you have any concerns about dropping the parens?

Also, I wouldn't bother mentioning `<script marko-init>` in the docs since that is deprecated.

patrick-steele-idem:
On a related note, we might have issues when we update the parser to handle commas so we probably need to provide a helper to get the raw string inside the open tag.

scttdavs:
Sure that looks like a simple change. I'll have time tonight to work on it
some more

scttdavs:
@patrick-steele-idem What do you think of the above implementation in the last commit? Instead of joining the attribute names for now (sounds like comma parsing will conflict with that later), I add the tag string to the node, so `el.tagString ` returns `"<import foo from './foo' />"`, and we get all we need from that. Is there a better way of doing this? Still feeling my around :)

patrick-steele-idem:
`tagString` would have the problem that it would be different for concise vs non-concise unless we always stripped off the the starting `<` and the ending `/>` or `>`. If we strip off those characters then I am okay with using `tagString`.

scttdavs:
@patrick-steele-idem any more comments, or is this ready to merge?

patrick-steele-idem:
It's ready to merge :) I am just finishing up the changes to support nested `<@foo>` tags and I will merge your changes very soon. We will then publish a new beta release. Thank you, again!

scttdavs:
great, thx! Also made one slight more optimization, couldn't help myself 🤓

PR #458, opened 2016-12-06
Remove direct require of test utils from within the compiler

Hesulan:
Fixes a regression which caused the compiler to `require()` both 'require-self-ref' and '~/test/util/patch-module'. This didn't affect the test suite (which already `require()`s both of those modules), but does throw errors when running `marko` as a module. For example:
```
npm install https://github.com/marko-js/marko
./node_modules/.bin/markoc ./template.marko
```
If `require-self-ref` is also installed:
```
Failed to compile "template.marko". Error: Error: Cannot find module '/path/to/node_modules/marko/test/util/patch-module'
```
Otherwise:
```
Failed to compile "template.marko". Error: Error: Cannot find module 'require-self-ref'
```

### DIFF
--- a/widgets/taglib/getRequirePath.js
+++ b/widgets/taglib/getRequirePath.js
@@ -1,6 +1,4 @@
 var resolveFrom = require('resolve-from');
-require('require-self-ref');
-require('~/test/util/patch-module');
 
 module.exports = function getRequirePath(target, context) {
     var resolvedTarget = resolveFrom(context.dirname, target);


Hesulan:
P.S.: The failing '[component-template-entry]' test was already like that - I didn't do it, I promise! :)

Hesulan:
I suspect it was an artifact from the `marko-widgets` merge, or perhaps used to quickly test something and someone forgot to remove it. If it actually did serve some purpose, it's definitely not the *right* way to do it - the compiler has no business touching anything under `~/test/`.

Hesulan:
https://github.com/marko-js/marko/commit/3010570502cf1c1020e382850645cc5b4ae55874#diff-3b089ba208a1607c7efd545409abc834

Hesulan:
On a related note, it might be a good idea to set up some sort of automated test to check that `marko` runs correctly as a node module, like `npm install https://github.com/marko-js/marko; ./node_modules/.bin/markoc template.marko`, to catch this type of regression.

PR #482, opened 2016-12-22
Fixed bug where widgets initialized before the runtime is loaded are lost

Hesulan:
The widget runtime uses a custom `.concat` method, but the native array `.concat` method returns a new array without modifying the original, which results in `window.$widgets` being an empty array and the init code being lost if the widgets are initialized before the runtime registers its custom method.


patrick-steele-idem:
Good catch. Thank you! I'll work on adding a test

PR #484, opened 2016-12-24
failing test for for-tag w/ inter-tag spaces

ramses0:
## Description

marko4 behaviour differs when using `<for ( x in y )>` and currently only supports `<for(x in y)>` (note space differences).

## Motivation and Context

I rarely expect programming languages (and less- templating languages) to be whitespace sensitive.  Whitespace sensitivity in this way requires projects to update all cases of `<for...` and likely would increase adoption / upgrade costs.

### DIFF
--- a/test/autotests/render/for-tag-spaces/expected.html
+++ b/test/autotests/render/for-tag-spaces/expected.html
@@ -1,0 +1,1 @@
+<div>red</div><div>green</div><div>blue</div>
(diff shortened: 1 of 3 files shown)


Hesulan:
It looks like a regex added to validate the `in` syntax doesn't recognize whitespace immediately following the `(`.

Fix: #485

PR #489, opened 2016-12-28
For each property should work with Map instances.

### LINKED ISSUE #477: for(key,value in myMap) should work with Map instances

austinkelleher:
Fixes #477


patrick-steele-idem:
Nice. Thank you!

PR #510, opened 2017-01-05
Fixes #421 - Deprecate w-bind

### LINKED ISSUE #421: Deprecate `w-bind`
`w-bind` → `widget` or `component`

See also #393 #395 #399 

__Update:__ we decided to remove `w-bind` and alternatives completely. If you want to bind a widget to the DOM nodes you need to name the file `widget.js`, `component.js` or embed an inline `<script>` tag in the template with a `module.exports`/`export default`.

austinkelleher:


patrick-steele-idem:
I think more changes are required. The `handleRootNodes` transform that is being used to do the implicit binding to all root elements is also internally adding the `w-bind` attribute to make everything work: https://github.com/marko-js/marko/blob/6aaf8679a88e53cef644582e51edca94517a51e4/widgets/taglib/TransformHelper/handleRootNodes.js#L202

patrick-steele-idem:
Looks good. Thank you for working on that!

PR #562, opened 2017-02-03
Fix for class attribute so empty arrays are null;

seangates:
This fix will ensure that an empty array does not output a an empty class attribute: `class=""`.

## Description
- modified helpers.js line 55 adds ` || null` when the array is empty (because a joined empty array is falsy);
- fixed tests to accommodate

## Motivation and Context
I was attempting to use arrays for class names because of the utility of concatenating them together was taken up by Marko. But, when the array was ever empty, I would have the `class=""` throughout my app, and my tests. This issue added unnecessary bloat. Therefore, the motivation for this fix.

## Checklist:

- [X] All new and existing tests passed.

### DIFF
--- a/runtime/helpers.js
+++ b/runtime/helpers.js
@@ -52,7 +52,7 @@ function classListHelper(arg, classNames) {
 function classList(classList) {
     var classNames = [];
     classListHelper(classList, classNames);
-    return classNames.join(' ');
+    return classNames.join(' ') || null;
 }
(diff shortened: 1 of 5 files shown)


mindeavor:
Just to clarify – you meant to say "because a joined empty array is falsey", correct?

seangates:
@mindeavor Noted. Fixed!

PR #571, opened 2017-02-08
Fixes #570 - Deprecate ref in favor of key.

### LINKED ISSUE #570: [v4] Rename "ref" to "key"
Currently, Marko supports "scoped IDs" in the form of a "ref":

```html
<div>
  <button ref="ok" type="button">OK</button>
  <button ref="cancel" type="button">Cancel</button>
</div>
```

A nested element/widget can be looked up by a ref:

```js
var okButton = this.getEl('button');
```

A "ref" attribute results in an ID attribute being assigned to the final element. For example:

```html
<div>
  <button id="w0-ok" type="button">OK</button>
  <button cancel="w0-cancel" type="button">Cancel</button>
</div>
```

[morphdom](https://github.com/patrick-steele-idem/morphdom) uses the "id" attribute as a "key" to match up elements during a diff/patch. Therefore, marko is using the "ref" attribute for two purposes:

- To assign a key for diffing/patching
- To assign a key for referencing nested elements/widgets

A "key" is the more general term ("scoped-id" is more appropriate, but too long) and "key" is already well understood to be used for matching up elements during diffing/patching.

As a late breaking change for the Marko v4 release, we want to rename "ref" to "key". We will introduce deprecation warnings, but we do not want to ship the "v4" release with support for "ref".

Thoughts? Concerns? Thanks in advance.

austinkelleher:


austinkelleher:
The CI build fail looks to be an intermittent failure.

PR #613, opened 2017-03-07
Updated Koa docs to version 2.0 signature

tinovyatkin:
Koa 2.0.1 is released as `latest` version, so, these docs needs to be updated. Also added gzip example to feature Koa + Marko streaming flexibility.

### DIFF
--- a/docs/koa.md
+++ b/docs/koa.md
@@ -1,28 +1,54 @@
-Koa + Marko
-=====================
+# Koa + Marko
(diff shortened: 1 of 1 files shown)


patrick-steele-idem:
This looks great. Would you be interested in providing a working koa app in the following repo: https://github.com/marko-js-samples/marko-koa

(I added you as a collaborator)

PR #625, opened 2017-03-17
Update README.md

krypton97:
Updated `counter.marko` example in `ReadME.md` file to use `onCreate()` instead of `constructor()`


austinkelleher:
Thanks!

Issue #621, opened 2017-03-15
[Question] how to increase default await timeout?


mauricionr:
## Question

How to increase the default await time out?

### Context

i'm getting a lot's of timeout from await tag, i'm using marko `3.4.2`

How the best way to handle this?(Handle the error, notify user and try again)
Is possible increase the await-tag timeout for all await-tags?

### Actual Behavior
Async fragment failed. Exception: Error: Async fragment timed out after `10000ms` at Timeout._onTimeout (/var/app/current/node_modules/async-writer/src/AsyncStream.js:143:32) at tryOnTimeout (timers.js:224:11) at Timer.listOnTimeout (timers.js:198:5)

Async fragment failed (await-reorderer). Exception: Error: Async fragment failed. Exception: Error: Async fragment timed out after ` 10000ms` at Timeout._onTimeout (/var/app/current/node_modules/async-writer/src/AsyncStream.js:143:32) at tryOnTimeout (timers.js:224:11) at Timer.listOnTimeout (timers.js:198:5) at AsyncStream.error (/var/app/current/node_modules/async-writer/src/AsyncStream.js:365:13) at Timeout._onTimeout (/var/app/current/node_modules/async-writer/src/AsyncStream.js:143:26) at tryOnTimeout (timers.js:224:11) at Timer.listOnTimeout (timers.js:198:5)

Thanks

Hesulan:
Use `<await(...) timeout=10000>`, where `10000` is the time in milliseconds, or `0` to disable the timeout entirely. It's currently not possible to change the default timeout for all await tags.

http://v3.markojs.com/docs/marko/async-taglib/

mauricionr:
I would like to change the default timeout for all await-tag's

Hesulan:
If anyone is interested in implementing this, here's where the default is currently hard-coded: https://github.com/marko-js/marko/blob/master/taglibs/async/await-tag.js#L149

PR #680, opened 2017-04-26
Costomize default await timeout

mauricionr:
## Description

Adding option to customize default timeout for await tags

## Motivation and Context

We use await a lot and we need to customize the default await timeout for all await tags and we think this will help us to fix the timeout problem

I've opened an issue asking about this thing #621 

## Checklist:

### DIFF
--- a/compiler/config.js
+++ b/compiler/config.js
@@ -26,9 +26,13 @@ if (g.__MARKO_CONFIG) {
         writeToDisk: true,
 
         /**
+         * Default timeout for await tag
+         */
+        awaitTimeout: process.env.MARKO_TIMEOUT || 10000,
+        /**
          * If true, then the compiled template on disk will assumed to be up-to-date if it exists.
          */
-        assumeUpToDate: process.env.MARKO_CLEAN != null || process.env.hasOwnProperty('MARKO_HOT_RELOAD') ? false : ( NODE_ENV == null ? false : (NODE_ENV !== 'development' && NODE_ENV !== 'dev')),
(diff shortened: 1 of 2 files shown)


austinkelleher:
This change looks good to me. Do you mind squashing your commits? Thanks.

mauricionr:
Squashed @austinkelleher 

Thanks :rocket:

austinkelleher:
Actually after given this some more thought, I'm not sure if this is the right approach for globally overriding the default for a component. I don't think the compiler config is the appropriate place for such configuration. Have you considered overriding the `<await>` tag? I'll create an example of this and post a link here.

austinkelleher:
Here's an example of overriding the `await` core tag:

https://github.com/marko-js-samples/override-core-tag

mauricionr:
@austinkelleher i will discuss with my team about override the `<await>` tag, thanks for the demo!

austinkelleher:
@mauricionr In the demo you'll see that I copied part of the `<await>` marko.json. This is my proposed solution for that: https://github.com/marko-js/marko/issues/686

mauricionr:
i saw @austinkelleher, really good proposal, and with this override, we will be able to override any core tag?

austinkelleher:
@mauricionr After speaking with @patrick-steele-idem, we decided that instead of my old approach where I overrode the tag renderer and copying the `<await>` tag's `marko.json` file, we can just override the `await` tag completely and use a transformer to set the `timeout` attribute.

*await-timeout-transformer.js*
```js
module.exports = function(el, context) {
    if (!el.hasAttribute('timeout')) {
        el.setAttributeValue('timeout', context.builder.literal(30000));
    }
}
```

The [override-core-tag](https://github.com/marko-js-samples/override-core-tag) project has been updated with the changes. 

I'm going to go ahead and close this issue. Please let me know if you have any other questions.

mauricionr:
Thanks @austinkelleher and @patrick-steele-idem simpler than the other aproach

PR #681, opened 2017-05-01
Fix encoding issue in core-tags documentation.

DylanPiercey:
## Description
Was reading through docs and found some html entities.

## Motivation and Context
Pretty sure they were not suppose to be there 😜.

## Checklist:

### DIFF
--- a/docs/core-tags.md
+++ b/docs/core-tags.md
@@ -496,9 +496,9 @@ _output.html_
 ```html
 <div>
     This is just one
-    &lt;span if(foo)&gt;
+    <span if(foo)>
         Hello ${THIS IS NOT VALID}!
-    &lt;/span&gt;
+    </span>
     big text block
 </div>
 ```


austinkelleher:
Thanks for the PR. This is actually the correct behavior. Using the `static-text` attribute, Marko ignores HTML tags. The Marko.js website is not rendering the output correctly though [here](http://markojs.com/docs/core-tags/#codemarko-bodycode). It should look like this:

```marko
<div>
    This is just one
    &lt;span if(foo)&gt;
        Hello ${THIS IS NOT VALID}!
    &lt;/span&gt;
    big text block
</div>
```

austinkelleher:
Also, I've opened up https://github.com/marko-js/markojs-website/issues/11 to track the Marko.js website issue.

DylanPiercey:
Gotcha, thanks!

PR #682, opened 2017-05-02
Don't set `serializedGlobals` properties to `false` after reading them.

Hesulan:
Currently, this:
```js
app.locals.foo = 'foo!';
app.locals.serializedGlobals = { foo: true };
```
Doesn't work, because [`component-globals-tag.js`](https://github.com/marko-js/marko/blob/master/components/taglib/component-globals-tag.js#L20) sets each property to `false` after reading them. If there's a reason for doing that, I can't find it, and removing that line fixes the problem.

### DIFF
--- a/components/taglib/component-globals-tag.js
+++ b/components/taglib/component-globals-tag.js
@@ -17,7 +17,6 @@ function getSerializedGlobals(outGlobal) {
                         serializedGlobals = {};
                     }
                     serializedGlobals[key] = value;
-                    serializedGlobalsLookup[key] = false;
                 }
             }
         }


austinkelleher:
Thanks for the PR @Hesulan. I agree that this does not look right. I'm not sure if there was an underlying reason for this, so I'm going to leave this open until either Patrick or Michael looks at it too. /cc @patrick-steele-idem @mlrawlings

mlrawlings:
@Hesulan @austinkelleher I'm not sure.

@Hesulan would you add a test case that would fail without this change?  

@patrick-steele-idem Was this intentional?  If so, we should add a test that would fail if that line was removed.

austinkelleher:
@Hesulan Just following up to see if you'll have a chance to add a test case and further describe this fix? Thanks!

patrick-steele-idem:
This PR has been opened for awhile. I'm not sure there is an issue here so I am closing. If you still think this is a problem please provide more details.

PR #683, opened 2017-05-03
Proposition: add docs for Marko + Huncwot integration

zaiste:
A piece of documentation describing how to use Marko alongside [Huncwot](https://github.com/zaiste/huncwot). I'd be happy to hear your feedback.

## Description
I'm building a library similar to Express and Koa called Huncwot. I'm planning to use Marko as its default (and only) view engine. Huncwot is inspired by Clojure's ring with routes as (pure) functions returning *data* describing responses. In this pull request I'm describing how to use Marko with Huncwot.

## Motivation and Context
Server-side integration example using Huncwot is simpler and requires less code for the same result as in Koa or Express.

## Checklist:
- [ ] I have added tests to cover my changes. (N/A)

### DIFF
--- a/docs/huncwot.md
+++ b/docs/huncwot.md
@@ -1,0 +1,23 @@
+# Huncwot + Marko
+
+See the [marko-huncwot](https://github.com/zaiste/marko-huncwot) sample
+project for a working example.
+
+## Installation
+
+    yarn add huncwot marko
+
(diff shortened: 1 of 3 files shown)


austinkelleher:
Thanks for the docs! We love seeing additional integrations with Marko.

PR #697, opened 2017-05-20
Publish marko with both a "debug" and a "non-debug" version

patrick-steele-idem:
This PR introduces a new "dist" directory that gets generated during a publish using babel and a set of babel plugins. Initially, we are only minifying property names, but we will eventually use this to introduce better error messages in the debug build and those error messages and warning messages will be stripped out for the "dist" build. The "dist" build will automatically get used if `NODE_ENV` is anything other than `undefined`|`development`|`dev` or if `MARKO_DEBUG` env variable is set.


seangates:
👍🏻

PR #699, opened 2017-05-23
Reference app should point to "marko-lasso" because "ui-components-pl…

tindli:
…ayground" does not contain lasso

## Checklist:

- [ ] My code follows the code style of this project.
- [ ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/docs/lasso.md
+++ b/docs/lasso.md
@@ -4,7 +4,7 @@ The [lasso-marko](https://github.com/lasso-js/lasso-marko) plugin for [Lasso.js]
 
 Lasso.js provides Marko custom tags for injecting JavaScript and CSS bundles, images and other resources.
 
-The sample [ui-components-playground](https://github.com/marko-js-samples/ui-components-playground) app demonstrates how to build a production-ready web application using Marko and Lasso.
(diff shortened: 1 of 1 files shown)


austinkelleher:
`ui-components-playground` is using `marko-starter` under the hood. `marko-starter` is using `lasso`. Perhaps this isn't clear enough. Right now, we aren't actively pushing people to use `marko-starter` apps for production, but in the future we plan to encourage this. What do you think @patrick-steele-idem @mlrawlings? Should we switch the link to the `marko-lasso` sample?

patrick-steele-idem:
Hiding the implementation details of the JavaScript module bundler/asset pipeline might be a good thing because I don't want developers to get distracted by "Webpack vs Lasso vs ...". The negative is that new developers might have a harder time understanding what is going on. With that said, I think I'm in favor of showcasing `ui-components-playground` since it has less code and it is easier to see `marko` in action.

@mlrawlings, thoughts?

mlrawlings:
Sorry I'm so late on this... I think I agree with @tindli in that we should link to the repo where `lasso` is explicitly used.  That doc is about integrating lasso and marko.  If you're there, you probably want to see how it works, not have it hidden away.

I probably wouldn't hurt to mention that `marko-starter` also uses lasso behind the scenes and doesn't require any configuration to get started.

PR #716, opened 2017-06-01
fixes #688 - add reference to browser.json in meta dependencies

### LINKED ISSUE #688: Marko doesn't check for require: <package> in browser.json
As long as I don't specify a `class {}` tag in single-file-component marko ignores `require` in the `browser.json`

## Reproducable with:
`browser.json`
```
{
  "dependencies": ["require: test", "./style.less"]
}
```
`index.marko`
```
<p>test</p>
```
Results in `index.marko.js` - **require is MISSING!**
```
marko_template.meta = {
    deps: [
      "./style.less"
  };
}
```

## Workaround
 Add `class {}` to the single-file-component to "allow require things"

`index.marko`
```
class {}
<p>test</p>
```
Results in `index.marko.js`
```
marko_template.meta = {
    deps: [
       {
          type: "require",
          path: "./"
        }
      "./style.less"
  };
}
```

## Solution:
Also single file components which dont need a `class` tag should allow to require packages.

mlrawlings:
## Checklist:

Regarding documentation, where should this be mentioned, on the components page under a section that's like "if you're using lasso..." or in the lasso integration doc?


austinkelleher:
This change looks good to me. I think we should support loading `browser.json` dependencies in webpack like we do in lasso. This shouldn't happen (very unlikely), but I think if has anyone switched from lasso to webpack and has any dangling `browser.json` files, this could technically break them.

mlrawlings:
While I would like it to work with webpack, I think that's going to cause more pain and confusion.  We've been saying Marko 4 means you no longer need `browser.json`, but the cases where you still need it would be Vanilla JS (Scripts) and Remaps (potentially conditional).  Both of these are features supported by lasso but not by webpack, at least not by default and would requires changes to the webpack configuration, but specific to those files.

I'm concerned if we support webpack we'll end up in a situation where it kinda works but doesn't really.

Issue #690, opened 2017-05-11
Possible bug when <style> tag is at the root


patrick-steele-idem:
Given the following template:

```marko
<style> 
#search-guide { 


} 
</style> 
<div id="search-guide"> 
  $ console.log( 'search-guide::state - ', state ); 
  <div> 
    <!-- state.len: ${ state.len } --> 
  </div> 

    <button on-click('pickSomething')>Pick Something!</button> 

    <hr> 

    Picked: ${ state.len } 
    Pick Something Else! 
    <!-- <button on-click('pickSomething')>Pick Something!</button> --> 
</div>
```

## From @ramses0:

...this gets stuck in an infinite loop after the on-click, due to it getting stuck in the weeds with the `style` tag during    `$__rerender: function(isRerenderInBrowser) {` ... `                var targetEl = targetNode.firstChild;
                while(targetEl) {
                    var id = targetEl.id;`
yes, i know it's a terrible dom structure (working on it, eventually ;) ), but for some reason this seems like the only component which has this problem, in many other place i'm using the "dangling style tag" structure w/o issue.
moving the <style> into the <div> seems to make it not get confused ... i could try to get a smaller repro and file an issue if necessary but wanted to share as i found it.  

i didn't pursue it too much b/c i had a workaround (dump `<style>` inside a single top-level `<div>` tag per component instead of outside) and i'm planning in revamping all my styles eventually anyway

smulder:
I just came across this same bug to add a second to the vote... moving the style tag into the root div of the template fixed it for now.

austinkelleher:
Fixed in `marko@4.4.18`

PR #751, opened 2017-06-27
Fixes #749 #690 - Do not rely on root node having an ID on rerender a…

### LINKED ISSUE #749: Infinite loop into Component.js
https://github.com/marko-js/marko/blob/master/src/components/Component.js#L553

If Component don't have id that place become an infinite loop.

austinkelleher:


mlrawlings:
@austinkelleher Is this finished? The code change looks good, but the tests don't make any assertions.  Is the idea to just test that no errors are thrown?

austinkelleher:
@mlrawlings Before these changes, using multiple root nodes or style tag at the root would cause an infinite loop on re-render. The tests I wrote ensure that the code does not get stuck in an infinite loop. I'm not sure what the best way to provide meaningful assertion is without writing unit tests for `src/components/Component.js`.

mlrawlings:
Maybe rename `multiple-roots` to `rerender-multiple-roots` and add a comment to the end of the tests so that in the future it's more clear to anyone that might be looking at the test that it is just testing that it rerenders successfully?
```
// The rerender was successful: no error or infinite loop.  Test passes.
```

austinkelleher:
@mlrawlings Good suggestion. I made the changes and squashed. Let me know if you or @patrick-steele-idem have any more suggestions.

PR #756, opened 2017-07-11
Fixes #755 - Allow diffing of HTML, head, and body.

### LINKED ISSUE #755: Allow diffing of html, head, and body
We should allow the root page to be a component and diff the root HTML element, head, and body tags including the title tag inside of the head tag.

For example, calling `component.increment()` in the component below should cause the page to re-render and update the title from `Count: 0` to `Count: 1`:

```marko
lasso-page dependencies=input.browserDependencies lasso=input.lasso

class {
    onCreate () {
        this.state = {
            count: 0
        }
    }
    onMount () {
        window.app = this;
    }
    increment () {
        this.state.count++;
    }
}

<!DOCTYPE html>
html lang="en"
    head
        meta charset="UTF-8"
        title -- Count: ${state.count}
        lasso-head
    body

        div id="test"
        div id="mocha"
        div id="testsTarget"

        lasso-body

        init-components immediate

        lasso-slot name="mocha-run"

        browser-refresh
```

austinkelleher:
Relies on the following PRs:

https://github.com/patrick-steele-idem/warp10/pull/3
https://github.com/lasso-js/lasso/pull/211

### DIFF
--- a/src/components/Component.js
+++ b/src/components/Component.js
@@ -547,23 +547,31 @@ Component.prototype = componentProto = {
                 var fromEl;
 
                 var targetEl = targetNode.___firstChild;
-                while(targetEl) {
-                    var id = targetEl.id;
+                while (targetEl) {
+                    var nodeName = targetEl.___nodeName;
 
-                    if (id) {
-                        fromEl = fromEls[id];
-                        if (fromEl) {
-                            morphdom(
-                                fromEl,
-                                targetEl,
-                                globalComponentsContext,
-                                onNodeAdded,
-                                onBeforeElUpdated,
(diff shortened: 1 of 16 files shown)


mlrawlings:
I'm not sure doing this is a great idea, but it's probably the least surprising thing to support it.

patrick-steele-idem:
@mlrawlings 

> I'm not sure doing this is a great idea, but it's probably the least surprising thing to support it.

It definitely has caused some issues with folks trying marko and it was hard to explain that you need to separate out the page content from the dynamic component content. It's actually nice that the `document.title` will automatically update if the page re-renders in the browser.

PR #758, opened 2017-07-12
fixes #757 - `:key` attribute postfix

### LINKED ISSUE #757: Support :key postfix on attributes
## New Feature
Support `:key` postfix on attributes

### Description
`key` provides a scoped `id` within a component.  `:key` would support more easily referencing this `id`.

```marko
<button aria-label="Payment Status Info" aria-describedby:key="tooltip">
</button>
<div key="tooltip" role="tooltip">
    ...
</div>
```

```marko
<label for:key="name">Name</label>
<input type="text" key="name"/>
```

### Context

We already provide `for-key` for this purpose, but there are a number of `aria-` attributes that also reference dom ids.  We can already use `__component.elId`, but that's kinda supposed to be internal and it's ugly.

### Possible Implementation

Deprecate `for-key` in favor of `for:key` and `:key` on any other attribute.

### Open Questions

Nothing. Let's do it.

### Is this something you're interested in working on?

Yes :)

mlrawlings:
## Description

Deprecate `for-key`/`for-ref`/`w-for` in favor of `for:key`.  Add support for `:key` on any attribute.  

```marko
<label for:key="name">Name</label>
<input type="text" key="name"/>
```

```marko
<button aria-label="Payment Status Info" aria-describedby:key="tooltip">
</button>
<div key="tooltip" role="tooltip">
    ...
</div>
```

## Motivation and Context

Useful because many [ARIA relationship attributes](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques#Relationship_attributes) also need to reference the `id` of an element on the page

## Checklist:


mlrawlings:
@patrick-steele-idem The express tests are failing in node 4 saying the `debug` module is missing.  It looks like this was previously failing on master after https://github.com/marko-js/marko/commit/774ead0dcd01dd0c38a9138468f67766da9adaa9

There's no reference to `debug` in that commit, so it's possible it's a downstream dependency and not related to the code, but I haven't looked too closely.  Any thought on if that commit could be related?

mlrawlings:
The PR is now passing (since it is merged with master, which has the test fix).

src/components/taglib/TransformHelper/handleComponentKeyAttrs.js
```diff
@@ -0,0 +1,43 @@
...
+            }
+
+            el.removeAttribute(attributeName);
+        }
+    });
+
+    // END support for deprecated for attributes
+
+    el.attributes.forEach(attribute => {
+        if (keyPostfix.test(attribute.name)) {
```
austinkelleher: Using `.endsWith(...)` is much faster than testing the regex. I would recommend:

```js
const keySuffix = ':key';

module.exports = function handleComponentKeyAttrs() {
    ...
    el.attributes.forEach(attribute => {
        const attributeName = attributeName;
        if (attributeName !== keySuffix && attributeName.endsWith(keySuffix)) {
            // Make your dreams come true here...
        }
    });
}

File: src/components/taglib/TransformHelper/handleComponentKeyAttrs.js [hunk 2]
```diff
@@ -0,0 +1,43 @@
...
+            let unfixedName = attribute.name.replace(keyPostfix, '');
+            el.removeAttribute(attribute.name);
+            if (el.hasAttribute(unfixedName)) {
+                this.addError(`The "${attribute.name}" attribute cannot be used in conjuction with the "${unfixedName}" attribute. (` + (el.pos ? context.getPosInfo(el.pos) : context.filename) + ')');
```
austinkelleher: Looks like the `addError` function allows you to pass the position: https://github.com/marko-js/marko/blob/master/src/compiler/CompileContext.js#L284

PR #762, opened 2017-07-14
Fixes #761 - component.elId() does not work on the server

### LINKED ISSUE #761: component.elId() does not work on the server
The following component will render fine in the browser, but on the server the user will get an error (`TypeError: component.elId is not a function`):

```marko
class {}
<div>
  <input key="checkbox"/>
  <label for=component.elId('checkbox')/>
</div>
```

patrick-steele-idem:


test/autotests/render/component-elId/expected.html
```diff
@@ -0,0 +1 @@
+Hello World!
```
austinkelleher: Assertion is incorrect causing the tests to fail.

src/components/ServerComponent.js
```diff
@@ -52,6 +52,20 @@ class ServerComponent {
...
         return this.___state;
     }
+
+    elId(scopedId, index) {
+        var id = this.id;
+
+        var elId = scopedId != null ? id + "-" + scopedId : id;
+
+        if (index != null) {
+            elId += "[" + index + "]";
```
austinkelleher: Maybe add a test where `index` is passed to?

test/autotests/render/component-getElId/expected.html
```diff
@@ -0,0 +1 @@
+Hello World!
```
austinkelleher: Assertion is incorrect causing the tests to fail.

patrick-steele-idem:
Merged

PR #765, opened 2017-07-15
Infinite loop fix

westtrade:
I'am not completely sure if this is right solution, but if fromNextSibling = curFromNodeChild - in that place, in line 226 we have curFromNodeChild = fromNextSibling; - that's the same of curFromNodeChild = curFromNodeChild - and then we have infinitee loop, which I caught in my experiments with marko components

### DIFF
--- a/src/morphdom/index.js
+++ b/src/morphdom/index.js
@@ -171,7 +171,7 @@ function morphdom(
 
                                             var curToNodeChildNextSibling = curToNodeChild.___nextSibling;
                                             if (curToNodeChildNextSibling && curToNodeChildNextSibling.id === curFromNodeKey) {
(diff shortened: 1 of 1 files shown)


austinkelleher:
Hey @westtrade, thanks for the PR! Can you add a failing test case? Your test would go in `test/autotests/morphdom`

patrick-steele-idem:
Thanks for investigating this @westtrade! Your change does prevent the infinite loop, but it looks like the incorrect output is produced. After investigating I am able to reproduce the infinite loop in a test case and am working on a proper fix.

patrick-steele-idem:
I investigated some more and removed a problematic optimization that was not really optimization. This resolved the infinite loop issue.

PR #773, opened 2017-07-24
updating location of bindComponent helper

fierysunset:
Location was missing a `src/` since the structure was updated in https://github.com/marko-js/marko/commit/a602f3cd466ec14468cd8a90e7334299077c329c

### DIFF
--- a/src/components/taglib/component-tag.js
+++ b/src/components/taglib/component-tag.js
@@ -16,7 +16,7 @@ module.exports = function codeGenerator(el, codegen) {
 
     var bindComponentVar = context.addStaticVar('marko_bindComponent',
         builder.require(
-            builder.literal('marko/components/taglib/helpers/bindComponent')));
+            builder.literal('marko/src/components/taglib/helpers/bindComponent')));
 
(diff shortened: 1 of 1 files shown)


patrick-steele-idem:
Good catch. However, we don't want to hard code the `src/` or `dist/` directory into the path since that will vary based on the `NODE_ENV`. Can you please update your PR to use the following code?:

```js
var bindComponentVar = context.importModule(
  'marko_bindComponent',
  'marko/components/taglib/helpers/bindComponent');
```

Thanks!

fierysunset:
@patrick-steele-idem Thanks! I didn't actually look deeper into the value of `marko_bindComponent` since we just updated the string directly in my node modules.

patrick-steele-idem:
Thank you! If you want to take a stab at a test case that would also be greatly appreciated.

fierysunset:
@patrick-steele-idem Gotta finish marko4 migration first :)

patrick-steele-idem:
No worries! I'll work with @mlrawlings to add a test case after merging. Thanks again.

PR #791, opened 2017-08-08
Fixes #790 - Class tag should not allow nested body content.

### LINKED ISSUE #790: Class tag should not allow nested body content
The following should throw an error:

```marko
class {}
  <div/>
```

austinkelleher:


mlrawlings:
Looks good.  

Since this change is to make it more clear that an unintended indent is the issue... 
Do we have easy access to the way the tag was nested?  If so, the error could add some more content, only in that case (this would be generic for any tag that does not allow body content). 

```
The "class" tag does not allow nested body content.  Check that you don't have accidentally indented content.
```

PR #792, opened 2017-08-08
Simplify a condition judgement in components-jquery

ulivz:
## Description
In the client side, we are very cared about the size of core runtime library size. So there is necessary to simplify the code as long as we can, of course it should be in the premise of not impacting the existing logic and performance. So i made this PR.

## Checklist:

- [ ] I have updated/added documentation affected by my changes. (NO NEED)
- [ ] I have added tests to cover my changes. (NO NEED)


austinkelleher:
Looks good to me. Thanks.

PR #794, opened 2017-08-08
Code simplification and size-reducing in client side

ulivz:
## Summary
Similar to https://github.com/marko-js/marko/pull/792, 
to simplify and clarify the code in client side.

### DIFF
--- a/src/components/ComponentDef.js
+++ b/src/components/ComponentDef.js
@@ -37,13 +37,7 @@ ComponentDef.prototype = {
      * so that we can instantiate nested components before their parents.
      */
     ___addChild: function (componentDef) {
-        var children = this.___children;
-
-        if (children) {
-            children.push(componentDef);
-        } else {
-            this.___children = [componentDef];
-        }
+        (this.___children || (this.___children = [])).push(componentDef);
     },
     /**
      * This helper method generates a unique and fully qualified DOM element ID
--- a/src/components/beginComponent.js
+++ b/src/components/beginComponent.js
@@ -5,15 +5,10 @@ const hasRenderBodyKey = Symbol.for("hasRenderBody");
 
 function isInputSerializable(component) {
     var input = component.___input;
-    if (!input) {
-        return true;
-    }
-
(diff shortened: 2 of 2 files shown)


src/components/ComponentDef.js
```diff
@@ -37,13 +37,7 @@ ComponentDef.prototype = {
...
      */
     ___addChild: function (componentDef) {
-        var children = this.___children;
-
-        if (children) {
-            children.push(componentDef);
-        } else {
-            this.___children = [componentDef];
-        }
+        (this.___children || (this.___children = [])).push(componentDef);
```
patrick-steele-idem: This change might have an impact on performance since it is hot code. My concern is that creating an array in one operation and then pushing to it in another operation might be slower. It is worth running some benchmarks to compare performance.

src/components/beginComponent.js
```diff
@@ -5,15 +5,10 @@ const hasRenderBodyKey = Symbol.for("hasRenderBody");
 
 function isInputSerializable(component) {
     var input = component.___input;
-    if (!input) {
-        return true;
-    }
-
-    if (input[hasRenderBodyKey] === true || input.renderBody !== undefined) {
+    if (input && (input[hasRenderBodyKey] === true || input.renderBody !== undefined)) {
```
patrick-steele-idem: A minifier will completely rewrite this code so I'm thinking it is best to keep the code readable. Here's the output of Google Closure Compiler on the original code (formatted for readability):

```js
function isInputSerializable(a) {
    a = a.___input;
    return !a || !0 !== a[hasRenderBodyKey] && void 0 === a.renderBody ? !0 : !1
};
```

ulivz: nice, I support your point.

src/components/ComponentDef.js
ulivz: This code style is followed the existing code. Sorry, I should have had more carefulness for the performance to contribute code.

ulivz:
Let me close this PR.

PR #795, opened 2017-08-08
Added `escapeAtTags` option to compiler in order to render `<@tags>` as they are;

oxala:
## Description
When `escapeAtTags` options is being passed to the compiler - it will escape `<@whatever></@whatever>` tags with `<at_whatever></at_whatever>`.

## Motivation and Context
During tests, if we isolate tested component from its nested dependencies, we may find ourselves in place where we have `<@whatever></@whatever>` tag sitting not within a custom component element.

*Desired result*: if we're isolating dependencies during tests, we want to render the transclusion code as it is.

## Checklist:
- [ x ] My code follows the code style of this project.
- [ - ] I have updated/added documentation affected by my changes.
- [ x ] I have read the **CONTRIBUTING** document.
- [ x ] I have added tests to cover my changes.
- [ x ] All new and existing tests passed.

### DIFF
--- a/src/compiler/CompileContext.js
+++ b/src/compiler/CompileContext.js
@@ -137,6 +137,7 @@ class CompileContext extends EventEmitter {
         this.compilerVersion = this.options.compilerVersion || markoPkgVersion;
         this.writeVersionComment = writeVersionComment !== 'undefined' ? writeVersionComment : true;
         this.ignoreUnrecognizedTags = this.options.ignoreUnrecognizedTags || false;
+        this.escapeAtTags = this.options.escapeAtTags || false;
 
         this._vars = {};
         this._uniqueVars = new UniqueVars();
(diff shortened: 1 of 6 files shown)


src/compiler/CompileContext.js
```diff
@@ -137,6 +137,7 @@ class CompileContext extends EventEmitter {
         this.compilerVersion = this.options.compilerVersion || markoPkgVersion;
         this.writeVersionComment = writeVersionComment !== 'undefined' ? writeVersionComment : true;
         this.ignoreUnrecognizedTags = this.options.ignoreUnrecognizedTags || false;
+        this.escapeAtTags = this.options.escapeAtTags || false;
```
mlrawlings: Is this used anywhere?  Did you intend to use it in the `if` statements below?  Or is it just to have easy access from elsewhere?

oxala: Oh, good catch. Sorry about that.
Let me actually use `this.escapeAtTags` on lines 464 and 479.

austinkelleher:
Published `marko@4.4.24` with these changes 🎉 🚀

Issue #815, opened 2017-08-24
Support Updating $global when setting new input


DylanPiercey:
## New Feature
Add ability to override the current `out.global` by setting new `$global` property on input.

### Description
Currently whenever you wish to override any global values you must scrap the existing component and render a new one. Being able to update globals on the fly has some benefits, and is useful for being able to update non static globals. Some examples of non static globals are: Request information (pathname, query, etc), storage (eg if using redux style shared stores passed from globals) and I'm sure there are more instances where this could come in handy.

### Context
I ran into this issue whilst developing a Rill middleware for marko and found that there was no convenient way of updating the request data in a component without discarding it.

### Possible Implementation
AFAIK this is a simple change and I will create a PR along with this issue.

### Open Questions
It may be worth considering if `___global` should be exposed, or setting `input` with `$global` is best. I personally vote for setting `input`.

### Is this something you're interested in working on?
Yes

PR #816, opened 2017-08-24
Add ability to update globals by setting new input.

DylanPiercey:
Add ability to override the current out.global by setting new $global property on input.

## Description
The change here is small but allows updating the global context without discarding the marko component or using hidden properties like `___global`.

## Motivation and Context
Fixes https://github.com/marko-js/marko/issues/815

## Checklist:

### DIFF
--- a/docs/components.md
+++ b/docs/components.md
@@ -483,7 +483,7 @@ this.setStateDirty('numbers');
 
 ### `this.input`
 
-The current input for the component.  Setting `this.input` will result in the component being re-rendered.
+The current input for the component. Setting `this.input` will result in the component being re-rendered. If a `$global` property is set the `out.global` will also be updated during the re-render, otherwise the existing `$global` is used.
 
 ## Variables
 
--- a/src/components/Component.js
+++ b/src/components/Component.js
@@ -437,6 +437,9 @@ Component.prototype = componentProto = {
 
         if (this.___input === undefined) {
             this.___input = newInput;
+            if (newInput.$global) {
+                this.___global = newInput.$global;
+            }
         }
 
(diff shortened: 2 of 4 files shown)


patrick-steele-idem:
I'm wondering if the following makes more sense:

```js
component.input = { foo: 'bar' };
component.global = { pathname: '/test' }
```

I don't have a strong preference. There is definitely some value in supporting `$global` on `input` to be consistent with how the globals are initially provided, but just wanted to throw this out there as an alternative.

DylanPiercey:
@patrick-steele-idem I personally think consistency with the current `render` api is probably best. As a user if anything I would expect `component.$global = ...` over `component.global = ...`.

Still I know I personally was surprised that `input.$global` only worked on the initial render.

patrick-steele-idem:
Thoughts @mlrawlings @austinkelleher ?

patrick-steele-idem:
@DylanPiercey One last thought/request. We should probably put this in the docs. Would you mind add some details in the following section as part of this PR?: https://github.com/marko-js/marko/blob/master/docs/components.md#thisinput

DylanPiercey:
@patrick-steele-idem, just updated the documentation as you mentioned. Let me know if the wording makes sense as the coffee hasn't kicked in yet 😜.

DylanPiercey:
How did coverage decrease 🤔

austinkelleher:
I've seen the coverage issue before. It's usually a coveralls issue. I'm going to go ahead and merge this PR. Thanks!

austinkelleher:
Published `marko@4.4.27` with this change. Thanks again!

PR #831, opened 2017-09-04
fix undefined reference error in Component.js

wujjpp:
## Description
before access newInput.$global， we should judge if it is undefined or not

### DIFF
--- a/src/components/Component.js
+++ b/src/components/Component.js
@@ -437,7 +437,7 @@ Component.prototype = componentProto = {
 
         if (this.___input === undefined) {
             this.___input = newInput;
-            if (newInput.$global) {
+            if (newInput && newInput.$global) {
                 this.___global = newInput.$global;
             }
         }


austinkelleher:
The only time I think this would happen is if you were setting `component.input = null`. Is that your use case?

wujjpp:
@austinkelleher YES

Scenario:

SSR

// file structure


app-sorter was rendered at server side, by it's state. The state data is from input, at client side app-sorter was manage state by it's self and emit changed event to parent component.  parent component handle event for fetching data and set it's state for other components(such as list, paginator and etc)

the problem was: after mount at client side, the error occurs.

components/app/index.marko


components/app/component.js


components/app-sorter/index.marko


components/app-sorter/component.js

wujjpp:
@austinkelleher YES

Scenario:  we receive `sorter` data from query string and initialize app-sorter component.

1. server router
[https://github.com/wujjpp/marko-starter-kit/blob/test-issue/src/routes/test-issue/index.js#L5-L18](url)


2. components/app/index.marko
https://github.com/wujjpp/marko-starter-kit/blob/test-issue/src/routes/test-issue/components/app/index.marko#L1-L5


3. components/app-sorter/index.marko
[https://github.com/wujjpp/marko-starter-kit/blob/test-issue/src/routes/test-issue/components/app-sorter/index.marko#L1-L8](url)


4. components/app-sorter/component.js
[https://github.com/wujjpp/marko-starter-kit/blob/test-issue/src/routes/test-issue/components/app-sorter/component.js#L1-L18](url)


5. undefined error


6. normal

austinkelleher:
Thanks for taking the time to investigate this issue and submitting a fix! Published `marko@4.4.28` with this fix.

Issue #832, opened 2017-09-07
Support SSR mounting without <body> tag.


DylanPiercey:
### Description
Currently when you use SSR and do not have a `<body>` tag the component will not initialize in the browser.

### Context
Sometimes it is useful for the actual document wrapper to be handled outside of Marko and only having Marko elements inside of <body>.

### Possible Implementation
See https://github.com/marko-js/marko/pull/833.

### Is this something you're interested in working on?
Yes

PR #833, opened 2017-09-07
Support mounting server side code without <body> tag in marko template

DylanPiercey:
## Description
Resolves https://github.com/marko-js/marko/issues/832.

This change adds `componentGlobals`, `initComponents` and `awaitReorderer` to templates if the `<body>` element is missing to allow for client side mounting.

This has been done by adding a new `meta` property to components that have a `<body>` tag at build time. Then at runtime after the component runs an `afterRender` hook that has been added which in the case of `runtime/html/template.js` will check for the meta property and add initialization code if it was missing.

The reason it was implemented this way is because adding the init components at compile time created a lot of extra bulk on each template. However when you do have a body tag it is best if the initialization code is added inside the element.

## Motivation and Context
Currently when rendering from the server Marko will only add initialization code if a `<body>` tag is present.

Sometimes it is useful for the actual document wrapper to be handled outside of Marko and only having Marko elements inside of the body. However if you do this you loose many of the benefits of SSR currently.

## Checklist:

### DIFF
--- a/src/components/taglib/body-transformer.js
+++ b/src/components/taglib/body-transformer.js
@@ -1,6 +1,8 @@
 'use strict';
 
 module.exports = function transform(el, context) {
+    context.setMeta('hasBody', true);
+
     let componentGlobalsNode = context.createNodeForEl('component-globals');
     el.prependChild(componentGlobalsNode);
 
--- a/src/runtime/html/Template.js
+++ b/src/runtime/html/Template.js
@@ -2,6 +2,9 @@
 var AsyncStream = require('./AsyncStream');
 var makeRenderable = require('../renderable');
 var stream = require('stream');
+var componentGlobals = require('../../components/taglib/component-globals-tag');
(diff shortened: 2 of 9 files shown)


src/components/taglib/body-transformer.js
```diff
@@ -1,6 +1,8 @@
 'use strict';
 
 module.exports = function transform(el, context) {
+    context.setMeta('hasBody', true);
```
DylanPiercey: Need to figure out how to account for when the body is in a nested component.

mlrawlings:
Closing as I don't think this is the right approach.  We should have a way to emit a `before-end` event or something that fires before `</body>` is written, or before `out` ends if there is no `<body>`.  At that point, users of the API can add additional things to the output and Marko can use that to inject `componentGlobals`, `initComponents` and `awaitReorderer`.

We should discuss more in #832

PR #840, opened 2017-09-13
Key scope fix

mlrawlings:
## Description

- fixes scoped/keyed attributes with dashes causing compile-time error (`aria-described-by:scoped`)
- maintains behavior prior to the current beta with a `key` being used alongside `component.elId` or `component.getElId`

```marko
<label for=component.getElId('name')>Name</label>
<input key="name"/>
```

## Checklist:

### DIFF
--- a/src/compiler/Parser.js
+++ b/src/compiler/Parser.js
@@ -266,10 +266,9 @@ class Parser {
                 }
 
                 if (attrDef.rawValue) {
-                    let match = /^component\.(?:getE|e)lId\((?:"|')(.*)(?:"|')\)$/.exec(attrDef.rawValue);
-                    if (match) {
-                        context.data.imperativeComponentIds = context.data.imperativeComponentIds || [];
-                        context.data.imperativeComponentIds.push(match[1]);
+                    if (/^component\.(?:getE|e)lId\(.*\)$/.test(attrDef.rawValue)) {
+                        // TODO: add complain call here
+                        context.data.hasImperativeComponentIds = true;
                     }
                 }
 
--- a/src/compiler/Parser.js
+++ b/src/compiler/Parser.js
@@ -265,6 +265,14 @@ class Parser {
                     }
                 }
 
+                if (attrDef.rawValue) {
+                    let match = /^component\.(?:getE|e)lId\((?:"|')(.*)(?:"|')\)$/.exec(attrDef.rawValue);
+                    if (match) {
+                        context.data.imperativeComponentIds = context.data.imperativeComponentIds || [];
+                        context.data.imperativeComponentIds.push(match[1]);
+                    }
+                }
+
                 parsedAttributes.push(attrDef);
             });
         }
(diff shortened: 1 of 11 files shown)


src/compiler/Parser.js
```diff
@@ -265,6 +265,14 @@ class Parser {
                     }
                 }
 
+                if (attrDef.rawValue) {
+                    let match = /^component\.(?:getE|e)lId\((?:"|')(.*)(?:"|')\)$/.exec(attrDef.rawValue);
```
austinkelleher: Small edge case, but this will not match for template strings.

src/components/taglib/TransformHelper/assignComponentId.js
```diff
@@ -99,6 +99,12 @@ module.exports = function assignComponentId(isRepeated) {
                 el.setAttributeValue('id', this.buildComponentElIdFunctionCall(assignedKey));
             }
 
+            if (context.data.imperativeComponentIds &&  el.data.userAssignedKey !== false) {
+                if (context.data.imperativeComponentIds.includes(assignedKey.value)) {
```
austinkelleher: Just putting this here as a reminder than `Array.prototype.includes` is not supported in Node 4

src/compiler/Parser.js
mlrawlings: This only supports string literals at the moment.  Also, should walk the AST rather than using a regex here?

austinkelleher: Yeah I mentioned the string literal bit in my comment above. Additionally, I do think that it's important to walk the AST here too. Technically someone could assign a variable:

```marko
$ var myId = component.getElId('name')

<label for=myId>Name</label>
<input key="name"/>
```

mlrawlings:
> Also can you add a test for when the rawValue actually is component.getElId(...) and when it is component.elId(...)

[autotests/render/component-elId](https://github.com/marko-js/marko/blob/0968d7aaf8a13855be452e70af5887cc593b8f53/test/autotests/render/component-elId/template.marko)
[autotests/render/component-getElId](https://github.com/marko-js/marko/blob/0968d7aaf8a13855be452e70af5887cc593b8f53/test/autotests/render/component-getElId/template.marko)

patrick-steele-idem:
These changes look good, but my only concern is that `component.elId()` is not deprecated so we we will have to continue to support this old behavior of assigning an `id` attribute whenever a `key` attribute is used in a template that also makes use of `component.elId()`. How would we go about moving away from this behavior to something we want to long term? Is there a way to show a warning? I suppose assigning an `id` attribute is not so bad, but I feel like it would be surprising.

patrick-steele-idem:
In theory, we could deprecate `component.getElId()` and `component.elId()` in favor of one of the following:

- `component.uniqueId(scopedId)` 
- `component.uid(scopedId)`

This would allow us to deprecate the changes of assigning an `id` to keyed elements when using `component.elId()` or `component.getElId()`. Thoughts?

mlrawlings:
@patrick-steele-idem I think that's the right approach, but we need to talk about the naming of the function.  I've opened #843 for discussion.  In the meantime, let's merge this to preserve the previous behavior for those on the beta and we can add the deprecation before moving to a stable release

PR #849, opened 2017-09-17
Fixes #848 browser-refresh does not reload marko files if a static reference is held

### LINKED ISSUE #848: browser-refresh does not reload marko files if a static reference is held
### Context
`browser-refresh` does not hot reload marko templates. I'm testing the beta version but might affect generally available version as well.

### Expected Behavior

```
// module initialization
const template = require('./some-template.marko');

// later in code:
template.render(input, out);
```

If you modify `./some-template.marko` then I would expect the page to refresh and use the new template. However, the `template` render function produces the same output as before the change.

### Actual Behavior

The old version of the template is used after a page refresh triggered by `browser-refresh`. You have to restart the app to see the change.

### Possible Fix

The `src/hot-reload.js` module monkey matches the `t` function exported by `src/runtime/html/index.js`. However, due to a circular dependency, the original `t` is not yet set when `hot-reload.js` tries to replace it. Basically, the `t` function that is set by `hot-reload` is overridden by the real `t` function exported by `src/runtime/html/index.js`. The short-term fix to export the `t` function in `src/runtime/html/index.js` before anything is required.

<details><summary>Additional Info</summary>

### Your Environment

Using `marko@4.5.0-beta.2`

</details>

philidem:
The main change is to move the following block above any requires:

```javascript
exports.t = function createTemplate(path) {
     return new Template(path);
};
```

This avoids issue with circular dependency which broke the hot-reload code because the `t` function was not exported right away.

### DIFF
--- a/src/runtime/html/index.js
+++ b/src/runtime/html/index.js
@@ -1,8 +1,4 @@
 'use strict';
-require('../../');
-
-var AsyncStream = require('./AsyncStream');
-var Template = require('./Template');
 
 /**
  * Method is for internal usage only. This method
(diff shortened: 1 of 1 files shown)


philidem:
NOTE: The diff does a poor job of showing the change

PR #855, opened 2017-09-22
Fixes #854 - Make every .marko file a UI component

### LINKED ISSUE #854: Make every .marko file a UI component
Currently, if a `.marko` file does not have an inline component `class {}` and if there is no corresponding `component.js` or `component-browser.js` then it will not be made into a "full-blown Marko UI component". If `.marko` is not a full blown Marko UI then no keys are assigned (for DOM diffing/patching purposes) and there is no way to reference the it using an assigned key from the outer component. Not assigning keys and not mounting a UI component for some `.marko` files causes some issues and it results in inconsistencies.

Given that there is very little overhead in making something into a full-blown Marko UI component, we will be moving in that direction. Users of Marko should not notice any differences and there should be no breaking changes as long as you weren't doing something you shouldn't have been doing (see caveat below).

## Caveat

Currently, if you have a UI component defined for an `index.marko` and let's say it makes use of other templates that are not full-blow UI components. If keys are assigned in the other `.marko` files then they would become scoped in the UI component associated with the main `index.marko`. With the change to make every `.marko` file a full-blow UI component, the key will always be scoped to the UI component associated with the current `.marko` file. We believe the updated behavior is the correct behavior and the current behavior was never promoted or encouraged in our docs.

patrick-steele-idem:
## Description
See: https://github.com/marko-js/marko/issues/854

## Checklist:

- [ ] My code follows the code style of this project.


src/compiler/ast/Node.js
```diff
@@ -319,14 +319,14 @@ class Node {
...
-    _normalizeChildTextNodes(context, forceTrim) {
+    _normalizeChildTextNodes(context, forceTrim, forceTrimStartEnd) {
         if (this._childTextNormalized) {
             return;
         }
 
         this._childTextNormalized = true;
 
-        var trimStartEnd = this._trimStartEnd === true;
+        var trimStartEnd = forceTrimStartEnd === true || this._trimStartEnd === true;
```
austinkelleher: Is this change necessary? It looks like it isn't being used (commented out in `src/components/taglib/TransformHelper/handleRootNodes.js`).

src/compiler/config.js
```diff
@@ -65,7 +65,13 @@ if (g.__MARKO_CONFIG) {
         /**
          * Whether <@tags> should error when are used outside a custom component.
          */
-        escapeAtTags: false
+        escapeAtTags: false,
+
+        /**
+         * Controls whether or not a key should be assigned to all HTML
+         * and csutom tags at compile-time. The default is `true`
```
austinkelleher: `csutom` -> `custom` typo

src/components/beginComponent.js
```diff
@@ -34,6 +34,13 @@ module.exports = function beginComponent(componentsContext, component, isSplitCo
         return componentDef;
     }
 
+    if (isImplicitComponent === true) {
```
austinkelleher: Thoughts on moving this conditional above the one that is currently above it? Micro-optimization, but there's probably no sense in doing the bitwise operations above if this is `true`.

src/components/taglib/TransformHelper/assignComponentId.js
```diff
@@ -19,24 +19,7 @@ module.exports = function assignComponentId(isRepeated) {
...
-
-    if (!this.hasBoundComponentForTemplate()) {
-        // We are assigning a component ID to a nested component in a template that does not have a component.
-        // That means we do not have access to the parent component variable as part of a closure. We
-        // need to look it up out of the `out.data` map
-        if (!context.isFlagSet('hasComponentVar')) {
-            context.setFlag('hasComponentVar');
-
-            var getCurrentComponentVar = context.importModule('marko_getCurrentComponent',
-                this.getMarkoComponentsRequirePath('marko/components/taglib/helpers/getCurrentComponent'));
```
austinkelleher: Can `marko/components/taglib/helpers/getCurrentComponent.js` be deleted now? I don't think it's being used anymore.

src/components/taglib/TransformHelper/convertToComponent.js
```diff
@@ -92,7 +99,7 @@ module.exports = function handleComponentBind(options) {
     if (rendererModule) {
         if (rendererModule.inlineId) {
             markoComponentVar = rendererModule.inlineId;
-        } else {
+        } else if (!isImplicitComponent) {
```
austinkelleher: It's not possible for this to fall into an `else` case right? Otherwise `markoComponentVar` would be `undefined`. Maybe that's intended.

test/autotests/markoc/single-template-write-version-comment/test.js
```diff
@@ -1,15 +1,11 @@
...
 }
 
 exports.test = function(helpers) {
     helpers.spawnSync(['template.marko']);
 
     var compiledFile = helpers.readSync('template.marko.js').toString();
-    var expectedFile = _appendMarkoVersionComment(helpers.readSync('expected.js'));
     compiledFile = compiledFile.replace(/marko\/dist\//g, 'marko/src/');
-    expect(compiledFile).to.deep.equal(expectedFile);
+    helpers.compare(removeMarkoVersionComment(compiledFile), '.js');
```
mlrawlings: In any other case, I would say stripping the comment is fine, but isn't this test supposed to be asserting the existence of the comment?

test/autotests/async-render/components-await-title/expected.html
```diff
@@ -1 +1 @@
-<html><head><title>Welcome Frank</title></head><body><!--M#s0--><div>Hello</div><!--M/s0--><script>(function(){var w=window;w.$components=(w.$components||[]).concat({"w":[["s0",0,{},{"f":1}]],"t":["/marko-test$1.0.0/autotests/async-render/components-await-title/components/hello/index.marko"]})||w.$components})()</script></body></html>
\ No newline at end of file
+<html><head><title>Welcome Frank<script>(function(){var w=window;w.$components=(w.$components||[]).concat({"w":[["s2",0,{},{"f":1}]],"t":["/marko-test$1.0.0/autotests/async-render/components-await-title/components/hello/index.marko"]})||w.$components})()</script></title></head><body><!--M#s2--><div>Hello</div><!--M/s2--></body></html>
```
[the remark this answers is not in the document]
patrick-steele-idem: I've not noticed that was happening until now, but that is not going to be valid. I'll investigate.

test/autotests/components-browser-deprecated/widget-split-into-multiple-templates/components/app-stateful-button/index.js
```diff
@@ -1,79 +0,0 @@
-module.exports = require('marko/legacy-components').defineComponent({
```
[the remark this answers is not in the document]
patrick-steele-idem: For Marko v4, how long do the warnings need to be on the next patch version before we move this out of beta?

test/autotests/markoc/single-template-write-version-comment/test.js
patrick-steele-idem: Perhaps. I will take a closer look.

src/components/taglib/TransformHelper/convertToComponent.js
patrick-steele-idem: It's okay if `markoComponentVar` is undefined. That is handled by `markoComponentVar || builder.literal({})` later in the file.

src/components/beginComponent.js
patrick-steele-idem: It's actually going to be more common for the if condition on line 32 to evaluate to `true` so it's better to keep the `if (isImplicitComponent === true)` lower. Whether or not a UI component is implicit only matters if it is a top-level UI component outside of UI component that will re-render in the browser.

src/components/taglib/TransformHelper/assignComponentId.js
patrick-steele-idem: Yes, we can delete it. Good catch.

Issue #893, opened 2017-10-15
Numeric keyed elements in some case will not matched with ___keyedElements into init-components-browser.js:120


westtrade:
### Marko Version: 4.5.0
![ice_screenshot_20171015-181430](https://user-images.githubusercontent.com/13632738/31586121-59dffc40-b1d5-11e7-927e-ff56f084396e.png)

### Temporary Fix for templates
As temporary fix, can use named keys
```
div.dropdown [
		class=component.class
		role=input.role
		onMouseenter("reactOnHover", true)
		onMouseleavo("reactOnHover", false)
		key="dropdownRoot"
	]
```

PR #894, opened 2017-10-15
Fix for #893 issue

westtrade:
In some case in this place, without this check, delete will remove necessary numeric keys. This is a possible fix for #893 issue

ezgif-4-5685c0c879

### DIFF
--- a/src/components/util-browser.js
+++ b/src/components/util-browser.js
@@ -64,7 +64,9 @@ function destroyNodeRecursive(node, component) {
         var key;
 
         if (component && (key = node.___markoKey)) {
-            delete component.___keyedElements[key];
+            if (node === component.___keyedElements[key]) {
+                delete component.___keyedElements[key];
+            }
         }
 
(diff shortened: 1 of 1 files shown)


patrick-steele-idem:
Thank you for filing the issue and creating this PR. I was able to reproduce the problem, but it was not simple since I needed to assign the same key to different elements in the template and those different elements needed to have different tag names. Here is the test case I added: https://github.com/marko-js/marko/tree/master/test/autotests/components-browser/event-handler-non-bubbling-rerender-el-mismatch

The problem was that we add elements to the keyed lookup during diffing, but we defer destroying until after DOM diffing is completed. Because destroying happens later, a destroyed node with a key that matches an incompatible (i.e., different tag name) would cause the new keyed element to be deleted from the keyed lookup. Your fix solves the problem. Thank you! 👍 

I've merging this PR, but I'm surprised you ran into the issue. New version published:

```
+ marko@4.5.1
```

PR #895, opened 2017-10-15
Fix in Component.js

westtrade:
```
Component.js:500 Uncaught TypeError: Cannot read property 'nextSibling' of null
    at Component.___forEachNode (Component.js:500)
    at Component.___detach (Component.js:491)
    at morphChildren (index.js:242)
    at morphEl (index.js:509)
    at morphChildren (index.js:281)
    at morphEl (index.js:509)
    at morphChildren (index.js:281)
    at morphEl (index.js:509)
    at morphChildren (index.js:281)
    at morphComponent (index.js:124)
```

### DIFF
--- a/src/components/Component.js
+++ b/src/components/Component.js
@@ -499,7 +499,7 @@ Component.prototype = componentProto = {
         for(;;) {
             var nextSibling = currentNode.nextSibling;
             callback(currentNode);
-            if (currentNode == endNode) {
+            if (currentNode == endNode || !nextSibling) {
                 break;
             }
             currentNode = nextSibling;


patrick-steele-idem:
Thank you for the PR Popov! I would really like to add test case for this since `___endNode` should never be `null`/`undefined`. It seems like something else might be going wrong earlier and this change is just masking the problem. Could you provide some more details so that we can add a test case? Thank you in advance!

jasonmacdonald:
@patrick-steele-idem  I will add that I hit this error quite a bit. Often it is when using the <include> tag where the renderer is dynamic (a passed in tpl) and the root element of the renderer may, or may not, have a key. At least that seems to be the case when I hit it. Sometimes one child has a key, and the next doesn't (often because of a mistake I have made lol), but that seems to trigger this error when trying to swap back-and-forth between renderers.

hedav85:
I also run in that issue ```currentNode is null``` at ```Component.js:500``` when using include tag.

My template code looks like:
```javascript
<div.list>
    <for(widget in widgets)>
        <if(widget.show)>
            <include(component.getWidgetTemplate(widget), {...}) key=widget.instanceId />
        </if>
    </for>
</div>
```
If I switch my widgets on/off everything works fine, but the last widget in the collection will generate this error. If I switch off (```show=false```) the widget then it will not be rendered and if I switch on the last widget it will not be rendered because of this error.
I can reproduce this always with the last element which should be rendered in the UI. If I switch widgets in between the first and the last shown it will works correctly.

If I surround the ```<inlcude/>``` tag with a ```<div>``` then it works like expected.
```javascript
<div>
    <include(...), {...}/>
</div>
```

mlrawlings:
I'm going to close this as it's just been sitting.  If this is still an issue and someone can open a PR with a [failing test case](https://github.com/marko-js/marko/blob/master/.github/CONTRIBUTING.md#adding-a-failing-test-case) we'll take another look.

PR #935, opened 2017-11-27
Switch to jsdom for tests.

DylanPiercey:
## Description
Switches test suite from phantomjs to jsdom. This decreases test run time from ~7:30 to ~1:30 on my machine. Also several (large) dev dependencies were able to be removed.

## Motivation and Context
Improve amount of time it takes to run tests.

## Checklist:


.travis.yml
```diff
@@ -1,8 +1,8 @@
 sudo: false
 node_js:
-- "4"
```
austinkelleher: I realize jsdom doesn't support Node 4, but Marko 4 still does. This means we will either have to strengthen our linting rules to ensure we aren't breaking Node 4, or this is a major version.

test/__runner__/browser/index.js
```diff
@@ -124,23 +117,25 @@ function generate(options) {
...
 
-function getCoverageFile(testsFile) {
-    return './.nyc_output/' + md5Hex(testsFile) + '.json';
+        return new Promise(function (resolve, reject) {
+            window.addEventListener('load', function () {
+                var runner = window.MOCHA_RUNNER;
+                runner.on('end', function () {
+                    cleanup();
+                    runner.stats.failures.length
+                        ? reject(new Error(''))
```
austinkelleher: Should we report the errors? Perhaps `.join` them?

test/__runner__/browser/index.js
```diff
@@ -124,23 +117,25 @@ function generate(options) {
...
         });
-    });
-}
 
-function getCoverageFile(testsFile) {
-    return './.nyc_output/' + md5Hex(testsFile) + '.json';
+        return new Promise(function (resolve, reject) {
+            window.addEventListener('load', function () {
+                var runner = window.MOCHA_RUNNER;
+                runner.on('end', function () {
```
austinkelleher: If  the `runner` never emits `end`, will the tests infinitely hang?

test/__runner__/browser/index.js
```diff
@@ -124,23 +118,35 @@ function generate(options) {
...
-            console.log(options.testsFile)
-            mochaPhantomJSOptions.coverageFile = getCoverageFile(options.testsFile);
-        }
+        return new Promise(function (resolve, reject) {
+            window.addEventListener('load', function () {
+                var runner = window.MOCHA_RUNNER;
+                runner.on('end', function () {
+                    if (shouldCover) {
+                        var coverageFile = getCoverageFile(options.testsFile);
+                        fs.writeFileSync(coverageFile, JSON.stringify(window.__coverage__));
```
austinkelleher: If `writeFileSync` throws, the tests will hang because the promise will never reject. Additionally, since this code is already async, why not just use `fs.writeFile` instead?

DylanPiercey: Good catch. I thought of switching it but it was just left over from copy pasting.

I’ll clean that up.

.travis.yml
DylanPiercey: Yeah, Michael and I were talking about this. I agree it should be under a major. We were debating if it would make sense to just release a Marko v5 (and start releasing mostly non breaking majors more often) or I may just switch back to jsdom 9. We can hash this out tomorrow morning though.

test/__runner__/browser/index.js
```diff
@@ -124,23 +118,35 @@ function generate(options) {
 
 function runTests(options) {
     return generate(options).then(generated => {
-        console.log(`Running ${generated.url} using mocha-phantomjs...`);
-        var mochaPhantomJSOptions = { useColors: true };
+        var cleanup = JSDOM(fs.readFileSync(generated.url, 'utf-8'), {
```
austinkelleher: Also this does not have to be sync either

Issue #939, opened 2017-11-30
Diff trying to remove last child in multi root components twice.


DylanPiercey:
### Marko Version: x.x.x
4.6.0

### Details
Error when diffing while using multiple top level elements.

### Expected Behavior
The diff would succeed.

### Actual Behavior
The diffing algorithm can cause a node to be removed twice. Once by the component when it is [destroyed](https://github.com/marko-js/marko/blob/c9e7410ae1be142ed5b20a20f094c748135061b6/src/components/Component.js#L256) and again by morphdom [when it removes all detached nodes](https://github.com/marko-js/marko/blob/c9e7410ae1be142ed5b20a20f094c748135061b6/src/morphdom/index.js#L538).

This causes an error because we can't remove an element from the DOM twice.

### Possible Fix
Check to see if the element has been removed by the component before morphdom attempts to remove it.

Here is [the pr](https://github.com/marko-js/marko/pull/940) to resolve the issue.

austinkelleher:
Fixed in `marko@4.7.1`

PR #896, opened 2017-10-16
node.parentNode maybe null if user remove it by manual, it will cause "Cannot read property 'removeChild' of null" error

wujjpp:
## Description
Add checking parentNode before call "node.parentNode.removeChild(node)"
if user remove the node by manual, it will cause error

## Screenshots (if appropriate):
image

PR #940, opened 2017-11-30
Fix issue with multiple top level elements being removed twice.

DylanPiercey:
## Description
Prevent attempted to remove dom elements twice (once at component level and once by morphdom).

## Motivation and Context
See https://github.com/marko-js/marko/issues/939 .

## Checklist:

### DIFF
--- a/src/morphdom/index.js
+++ b/src/morphdom/index.js
@@ -532,7 +532,7 @@ function morphdom(
             var componentToDestroy = node.___markoComponent;
             if (componentToDestroy) {
                 componentToDestroy.destroy();
-            } else {
+            } else if (node.parentNode) {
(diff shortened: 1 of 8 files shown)


Heziode:
This PR trying to solve the same problem that #896, isn't it ?

DylanPiercey:
@Heziode they are similar. However #896 is for supporting dynamically removing nodes and allowing the component to ignore that. This PR is specifically for resolving an issue where a component with multiple roots can be attempted to be removed twice by morphdom. Ultimately this solution is not ideal and there should be some restructuring to the diffing code to more adequately support this, however it is a quick and easy fix for the time being.

I'd like to consider merging both this and #896 with a few modifications.

Issue #938, opened 2017-11-30
Input field loses focus on input when wrapped in a layout


Hafas:
### Details
When putting an `input` element in a layout component (e.g. with `include`) and updating the state `on-input`, the `input`-Element loses focus each time.

### Expected Behavior
The `input`-Element shouldn't lose focus `on-input`. (see https://hafas.github.io/marko-input-bug/dist/ok-example.html - Example does not use a layout.)

### Actual Behavior
The `input`-Element loses focus (see https://hafas.github.io/marko-input-bug/dist/not-ok-example.html)

### Steps to Reproduce
Source to the examples, here: https://github.com/Hafas/marko-input-bug

DylanPiercey:
This was patched in #966.

PR #966, opened 2018-01-03
Issue 938

DylanPiercey:
## Description
Currently there is a part of the diffing algorithm that temporarily removes a component from the Dom that causes it to loose focus. This patch keeps track of the focused element and refocuses if needed.

## Motivation and Context
#938 

## Checklist:

### DIFF
--- a/src/morphdom/index.js
+++ b/src/morphdom/index.js
@@ -245,8 +245,13 @@ function morphdom(
                         }
 
                         // We need to move the existing component into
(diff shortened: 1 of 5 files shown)


Eldar-X:
This solution only partially solves the problem.
For example, I have a framework that adds some animation classes when you focus on input fields. It listens to the focus in event, and when my component states change, my fields flash

PR #989, opened 2018-02-16
Testing Improvements.

DylanPiercey:
## Description
* Setup utility to run jsdom tests in the same process without bundling.
* Cleanup require hooks in test suite.
* Test organization and cleanup.
* Remove redundant autotest utilities.
* Work towards running `components-browser` tests with hydration.
* Remove lasso as a dev dependency.

## Motivation and Context
Some bugs have been discovered around some hydration edge cases.
To improve the test coverage for hydration we want to have all of the `components-browser` tests run through a hydration step and a browser render step.

With the existing setup the process will run out of ram when bundling all of these tests (twice) because of the way lasso is caching things.

This PR has a WIP implementation to run the components browser tests in hydration mode using a new testing technique which removes the need for running browser tests through a bundler in order to run them.

This also makes the tests run much faster.

## Screenshots (if appropriate):
screen shot 2018-02-15 at 4 57 19 pm

## Checklist:


test/components-pages/fixtures-deprecated/getRenderedWidgets/tests.js
```diff
@@ -5,7 +5,7 @@ var markoWidgets = require('marko-widgets');
 describe(path.basename(__dirname), function() {
     it('should serialize widget config down to the browser', function() {
 
-        expect(window.simpleWidgets).to.equal(undefined);
+        expect(window.simpleWidgets).to.eql([]);
```
austinkelleher: Why did this change to an empty array?

mlrawlings: The tests used to run before the static component code:
https://github.com/marko-js/marko/blob/ecc7763f26d8cb25f13319b7f4bcea7c6171645a/test/components-pages/fixtures-deprecated/getRenderedWidgets/components/app-simple/index.js#L2

They now run after, but it doesn't change the meaning of the tests.  It's still saying the same thing: "expect there to be no `simpleWidgets` yet."

test/components-pages/fixtures/component-init-async/template.marko
```diff
@@ -1,5 +1,3 @@
-lasso-page dependencies=data.browserDependencies lasso=data.lasso
-
 import Promise from 'promise-polyfill';
```
mlrawlings: I notice you [upgraded this module](https://github.com/marko-js/marko/pull/989/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R101).  Is it even needed?

DylanPiercey: Yeah I think you are right. I thought I saw a test that was testing specifically for a poly filled promise but I guess I was wrong. I updated to PR to remove this dependency.

PR #1010, opened 2018-03-02
Hydrate fixes

mlrawlings:
FIX TRANSCLUDED
===============

Preserve included renderBodies
------------------------------
AsyncStream.js: add comment method
include-tag.js: wrap renderBodies in stateful components with comment markers to preserve content if the renderBody doesn't get serialized
morphdom/index.js: preserve include fragment markers
Component.js: skip non-element nodes for el and els (these might be the comment markers)
TEST: transclusion-include-split

Update legacy renderBodies to use the new method of preservation
----------------------------------------------------------------
components-transformer.js: w-body is now a plain include, no need to add additional preserve logic
renderer-legacy.js (51-156): ensure new body causes update
renderer-legacy.js (196): preserve legacy body variable
TEST: widget-stateful-preserve-body-from-split-component

Mount components in preserved renderBodies
------------------------------------------
renderer.js: add init code for components based on their template location, not render tree location
init-components-browser.js: walk into stateful trees to check for additional components
TEST: transcluded-component-from-non-rerender-root

FIX EVENTS
==========

Stateful components should rerender when mounting
-------------------------------------------------
beginComponent.js: don't check renderBody to determing rerender
TransformHelper/handleLegacyBind.js: for `index.js` set renderer to the component, so it is not marked as a split component
UPDATE_EXPECTATIONS: bind-component, component-include-attr

Ensure elements with events in split components have thier keys serialized
--------------------------------------------------------------------------
TransformHelper/index.js: add serialize key function which writes out `data-marko-key` attribute
TransformHelper/assignComponentId.js: use serialize key function
TransformHelper/convertToComponent.js: add variables for component type to context
TransformHelper/handleComponentEvents.js: serialize keys for direct event listeners
UPDATE_EXPECTATIONS: everything else

OTHER
=====

Call onCreate and onInput when mounting root rerender components
----------------------------------------------------------------
ComponentDef.js: call onCreate and onInput
(tests covered by changes to `onCreate` and lifecycle tests)

Fix issue with once-*() causing all events for hydrated split components to be handled only once
------------------------------------------------------------------------------------------------
event-delegation: string to boolean
TEST: event-handler-bubbling-once-split

Change the way input is removed from legacy components
------------------------------------------------------
defineWidget-legacy-browser.js


src/components/Component.js
```diff
@@ -544,14 +544,19 @@ Component.prototype = componentProto = {
...
+            el = el.nextSibling;
+        }
     },
 
     get els() {
         if ('MARKO_DEBUG') {
             complain('The "this.els" attribute is deprecated. Please use "this.getEls(key)" instead.');
         }
-        return getNodes(this);
+        return getNodes(this).filter(el => el.querySelector);
```
DylanPiercey: Needs to be changed to not use an arrow function.
Also we should probably change from a `el.querySelector` check to `e.nodeType === ELEMENT_TYPE_CONSTANT_FROM_SOMEWHERE`.

src/components/event-delegation.js
```diff
@@ -10,8 +10,12 @@ var listenersAttachedKey = '$MED' + runtimeId;
 function getEventFromEl(el, eventName) {
     var virtualProps = getMarkoPropsFromEl(el);
     var eventInfo = virtualProps[eventName];
+
     if (typeof eventInfo === 'string') {
         eventInfo = eventInfo.split(' ');
+        if (eventInfo[2]) {
+            eventInfo[2] = eventInfo[2] === 'true';
```
DylanPiercey: We could save some extra bytes if we serialize this to something other than the string 'true'.
Probably doesn't need to be addressed in this PR though.

PR #1016, opened 2018-03-08
Format + lint

mlrawlings:
Switch to `eslint`  and add `prettier`

Here's a link to the diff prior to running prettier:
https://github.com/marko-js/marko/compare/d7e529d401078817ee9cd4ab8887de9458956b4e


MaxMilton:
Congrats on this! Nice to see some modern tooling going into the project :+1: :star2: 

I just wish you had used a 2 space tab width... :sweat:

mlrawlings:
@MaxMilton We made the decision to just use the prettier defaults, however, it seems our [`.editorconfig`](https://github.com/marko-js/marko/blob/master/.editorconfig#L14) has a tab width set and prettier is picking that up.  

@marko-js/core Do we want to remove the `.editorconfig` and reformat or leave as is?   I
 don't have a preference myself.

As far as the single quotes, it looks like prettier [may default to that in 2.0](https://github.com/prettier/prettier/issues/4102).  We could go ahead and make the change now to avoid thrashing commits later... or we could wait.  Thoughts?

MaxMilton:
Ah, I was wondering where the config came from! I didn't realise prettier will pick up `.editorconfig` files when there's no explicitly set value.

The other option that looks like it will change is `trailingComma` defaulting to `es5` which would also make a big impact on this codebase. I say make the changes now but I'll leave that up to the core team :blue_heart:

PR #1019, opened 2018-03-13
Fix event collision with multiple runtimes.

CestDiego:
## Description

Verified that this way of getting the runtimeId was not working and this is a
way to have no problems with this.

## Motivation and Context

When injecting marko modules through AJAX (including another marko runtime) there is a collision on the `runtimeId` value. This fixes the event problems by forgetting about runtimeId and just setting the eventKey to a timestamp.

## Checklist:

* [x] My code follows the code style of this project.
* [x] I have updated/added documentation affected by my changes.
* [x] I have read the **CONTRIBUTING** document.
* [ ] I have added tests to cover my changes.
* [x] All new and existing tests passed.

### DIFF
--- a/src/components/event-delegation.js
+++ b/src/components/event-delegation.js
@@ -5,7 +5,7 @@ var getMarkoPropsFromEl = componentsUtil.___getMarkoPropsFromEl;
 
 // We make our best effort to allow multiple marko runtimes to be loaded in the
 // same window. Each marko runtime will get its own unique runtime ID.
-var listenersAttachedKey = "$MED" + runtimeId;
+var listenersAttachedKey = "$MDE" + runtimeId;
 
 function getEventFromEl(el, eventName) {
     var virtualProps = getMarkoPropsFromEl(el);
--- a/src/components/util-browser.js
+++ b/src/components/util-browser.js
@@ -1,6 +1,6 @@
 var markoUID = window.$MUID || (window.$MUID = { i: 0 });
 
-var runtimeId = new Date().getTime() + String(Math.random()).slice(2); // slice the non-decimal part of the random output
+var runtimeId = markoUID.i++;
 
 var componentLookup = {};
 
@@ -81,7 +81,7 @@ function nextComponentId() {
     // marko runtimes. This allows multiple instances of marko to be
     // loaded in the same window and they should all place nice
     // together
-    return "b" + markoUID.i++;
+    return "c" + markoUID.i++;
 }
 
 function nextComponentIdProvider() {
--- a/src/components/util-browser.js
+++ b/src/components/util-browser.js
@@ -1,5 +1,4 @@
 var markoUID = window.$MUID || (window.$MUID = { i: 0 });
-
 var runtimeId = markoUID.i++;
 
 var componentLookup = {};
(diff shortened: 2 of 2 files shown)


seangates:
@CestDiego Looks great to me. My only question is around the memory size needed to store lots of these IDs, but I'm sure it's pretty minimal at the end of the day. Do you think going from `0` to `15209789056946238358412916298` (example) is going to have any memory impact? (I doubt not, just asking because it's the first thing I thought of.)

DylanPiercey:
@seangates Originally I thought that this runtime ID was only used [here](https://github.com/marko-js/marko/blob/master/src/components/event-delegation.js#L124) to avoid attaching events multiple times.

However I realized that this is also used for the actual element and component ids which means that this could indeed have memory impact.

Currently trying to think around what the ideal solution is here.

---

For future reference:
What happened here is that previous to Marko 4.5 a different window variable was used to store and increment the current Marko runtime index. Because these different versions are using different counters they now think they have the same instance id (0) which is caused issues for @CestDiego.

src/components/util-browser.js
```diff
@@ -1,4 +1,5 @@
 var markoUID = window.$MUID || (window.$MUID = { i: 0 });
+
```
DylanPiercey: Would you mind removing this space from the PR?

CestDiego: done :)

src/components/util-browser.js
```diff
@@ -80,7 +80,7 @@ function nextComponentId() {
     // marko runtimes. This allows multiple instances of marko to be
     // loaded in the same window and they should all place nice
     // together
-    return "b" + markoUID.i++;
+    return "c" + markoUID.i++;
```
CestDiego: This allows the newest marko version to be compatible with every previous one. However we won't be able to make <4.4 and >4.5 versions compatible because the global object that held the runtimeId value gets deleted: https://github.com/marko-js/marko/blob/0f5e639775cb23a75fea5669e4a651ded537f6f7/src/components/init-components-browser.js#L219

src/components/event-delegation.js
```diff
@@ -5,7 +5,7 @@ var getMarkoPropsFromEl = componentsUtil.___getMarkoPropsFromEl;
 
 // We make our best effort to allow multiple marko runtimes to be loaded in the
 // same window. Each marko runtime will get its own unique runtime ID.
-var listenersAttachedKey = "$MED" + runtimeId;
+var listenersAttachedKey = "$MDE" + runtimeId;
```
CestDiego: This is also a fix so that the listeners attach to a new runtimeId in the newest Marko version. This makes the newest marko version compatible with all previous ones, but doesn't allow 4.5> to be compatible with 4.4

CestDiego:
Added some comments explaining the changes :) Thanks @DylanPiercey for your guidance on this issue.

@seangates the concern about memory use is totally valid and also the reason why we ended up going to the b=>c change. 

The change of the runtimeId started to just change the name of the Delegated Events variable, which only gets set once. It's in a commit no longer in this PR because it got force-pushed with the commit that changes the runtimeId instead.

DylanPiercey:
Released in 4.9.2 , thanks for your help @CestDiego!

PR #1032, opened 2018-04-09
allow loading components by lasso paths, but mark as deprecated

mlrawlings:
## Description

Prior to Marko 4.7, if you were using `lasso`, the following was how you would specify components that Marko could not be discovered through static analysis:

_browser.json_
```json
{
    "dependencies": [
        "require: path/to/component"
    ]
}
```

`require:` puts the required dependency into the bundle, but doesn’t load it.  And you need use require it using the lasso generated paths.  Marko currently uses lasso paths as component ids, but it shouldn’t rely on this and it also meant that we were very tied to lasso - certain features didn’t work well with other bundlers.

Marko 4.7 moves away from this coupling, but the intention was not to break existing behavior.  This PR adds back support for loading by lasso-specific urls, but logs a deprecation warning.

`lasso-marko@2.4.0` was released at the same time as Marko 4.7 and introduced the `marko-dependencies:` package type.  This should be used instead.

## Checklist:

* [x] My code follows the code style of this project.
* [ ] I have updated/added documentation affected by my changes.
* [x] I have read the **CONTRIBUTING** document.
* [ ] I have added tests to cover my changes.
* [x] All new and existing tests passed.

### DIFF
--- a/src/components/registry-browser.js
+++ b/src/components/registry-browser.js
@@ -1,4 +1,6 @@
+var complain = "MARKO_DEBUG" && require("complain");
 var defineComponent = require("./defineComponent");
+var loader = require("../loader");
 
 var registered = {};
 var loaded = {};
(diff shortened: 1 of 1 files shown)


mlrawlings:
This branch does not add any tests as we have removed `lasso` from our test suite, but it has been verified working in an existing application and all the existing tests pass. 

image

PR #1035, opened 2018-04-18
support array like targets in a <for(x in array)>

cameronbraid:
## Description

Currently the <for (x in arrayLike | status-var=loop)> tag works with array like objects - for example mobx observable array (https://mobx.js.org/refguide/array.html)

However the standard <for (x in arrayLike)> tag explicitly checks for Array.isArray which prevents the tag from being used for array like objects

This pull request enables the js:forEachHelper to iterate over array like objects, by not explicitly checking for Array.isArray.  It also brings it inline with helper-forEachWithStatusVar semantics by converting an object without a forEach property into a singleton array.

## Motivation and Context

It allows the standard <for(x in array)> to work seamlessly (without any special conversion to 'pure' arrays) with array like objects for example mobx observable array https://mobx.js.org/refguide/array.html

## Why is this change required? What problem does it solve?

It allows seamless integration of observable data models (like mobx) without the developer having to special case the use of observable arrays.

Without this change a marko user could take one of the following workarounds 

- use <for (x in array  **| status-var=loop**) which already works with array like objects
- convert the array like object into an array when using the standard for tag


mlrawlings:
Hey @cameronbraid, we're open to supporting other "array-like" objects in addition to arrays, but I don't think checking the presence of `forEach` is the best approach.  "Array-like" objects don't necessarily implement array methods (see `arguments`/`NodeList`).  Prior to ES2015, the typical way to check for an "array-like" has been to [check for a numeric length](http://2ality.com/2013/05/quirk-array-like-objects.html).  Something like:

```js
function isArrayLike(obj) {
  var length = obj != null && obj.length
  return typeof length === 'number' && length >= 0 && length % 1 === 0
}
```


However, in ES2015 the [iteration protocols](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols) were introduced for exactly this reason.  

It looks like the mobx arrays [do implement the iterable protocol](https://github.com/mobxjs/mobx/blob/d17e73cba664842048008cefbbd2243a6db1af30/src/types/observablearray.ts#L585), so this approach would work for them as well as Arrays, Generators, Maps, Sets, and other objects that conform to the protocol. 

However, because this is a newer protocol and we're still supporting browsers back to IE9, we'll need to have a fallback.  One thing that many third-party libraries (including mobx) do is use `"@@iterator"` in place of the Symbol where it's not supported.

```js
function iteratorSymbol() {
    return (typeof Symbol === "function" && Symbol.iterator) || "@@iterator"
}
```
https://github.com/mobxjs/mobx/blob/master/src/utils/iterable.ts#L8

Built in types won't have this though, so if we wanted full support, we'd need to check the length to do "array-like" duck typing.  The two primary types I'm aware of are `arguments` and `NodeList`, neither of which are likely to be used within a template.  

I'm thinking only supporting the iteration protocols (with the Symbol → String fallback) should be sufficient.  Thoughts?

/cc @marko-js/core

cameronbraid:
@mlrawlings yeah, I understand that, I was really just wanting to relax the requirement of using Array.isArray - which would solve the issue I currently face - which was that my templates were working when using for with status var, but when I remove the status var part of the expression the template no longer worked.

Then when I implemented the test case, I noticed that the 'for with loop var' helper has the check for .forEach so I added it into the 'standard for' helper to make them consistent.

cameronbraid:
this is now unnecessary in 4.14.0

PR #1060, opened 2018-06-21
Update dependencies and remove unused open

austinkelleher:
Fixes from `npm audit`. Remove `open` because I don't think it was being used.


mlrawlings:
The `npm run test-coverage` command uses it to auto open the coverage report because I'm lazy. 😂

It's not totally necessary, but we could also switch to [`opn`](https://www.npmjs.com/package/opn).

https://github.com/marko-js/marko/blob/76137568f20d047d7492f3d9f1543650e50b175a/package.json#L14

mlrawlings:
Also, did you regenerate the whole `package-lock.json`?  The build is failing because of some changes to the way `prettier` formats code.  

We either need to `npm run format` or only update these specific versions.

I think if you were to `git checkout master -- package-lock.json && npm i` it would update the package lock for only those packages.

PR #1062, opened 2018-06-21
Fix issue with CI failing due to check-fomat by running prettier.

austinkelleher:
It seems that the build keeps failing due to the npm `check-format` script. Running `prettier` on all of the files seems to fix the issue. We should consider adding `check-format` to the normal `test` script because it is not very clear why the tests fail in the CI otherwise.


.github/CODE_OF_CONDUCT.md
```diff
@@ -3,20 +3,20 @@
 This project adheres to the [eBay Code of Conduct](http://ebay.github.io/codeofconduct).
 By participating in this project you agree to abide by its terms.
 
-* Be friendly and patient.
```
austinkelleher: @mlrawlings I'm not sure if this was formatted this way for any reason, but prettier formatted it.

mlrawlings: Maybe I'm just seeing this is multiple projects, but it seems like prettier keeps switching between `-` and `*` for lists.  And it's driving me crazy.

Issue #1051, opened 2018-06-01
Uncaught TypeError: Cannot read property 'parentNode' of null


abiyasa:
### Details

When a page is rendered, we got an error on browser console which causing any component's browser-side-behaviour stop working at all:

```
Uncaught TypeError: Cannot read property 'parentNode' of null
    at init-components-browser.js:259
    at Array.forEach (<anonymous>)
    at Object.initServerRendered [as concat] (init-components-browser.js:222)
...
```

To isolate this bug, I made a repo so you can easily reproduce this bug: https://github.com/abiyasa/marko-dom-node-bugs

Steps to reproduce:

1. Clone https://github.com/abiyasa/marko-dom-node-bugs and do `yarn install`
1. `yarn start` to start server
1. open http://localhost:8080/clickable-expander or http://localhost:8080/clickable-expander-no-textual-display
1. Checkout the browser console

### Expected Behavior

The page should be rendered without error. The components (green expandable buttons) should be clickable which will reveal the hidden content.

### Actual Behavior

Errors appear on browser's console. Clicking the expandable buttons will do nothing.

### Possible Fix

The cause is unknown but if we remove the component.js, the error is also gone!

* Clone https://github.com/abiyasa/marko-dom-node-bugs and do `yarn install`
* `yarn start` to start server
* open http://localhost:8080/non-clickable-expander
* No error!

It seems something wrong on browser nodes or virtual DOM.

<details><summary>Additional Info</summary>

### Your Environment

* Environment name and version (e.g. Chrome 39, node.js 5.4): Node.js 8.11.x, Latest Chome & Firefox
* Operating System and version (desktop or mobile): OSX
* Link to your project: https://github.com/abiyasa/marko-dom-node-bugs

### Steps to Reproduce

1. Clone https://github.com/abiyasa/marko-dom-node-bugs and do `yarn install`
1. `yarn start` to start server
1. open http://localhost:8080/clickable-expander or http://localhost:8080/clickable-expander-no-textual-display
1. Checkout the browser console

### Stack Trace

```
Uncaught TypeError: Cannot read property 'parentNode' of null
    at init-components-browser.js:259
    at Array.forEach (<anonymous>)
    at Object.initServerRendered [as concat] (init-components-browser.js:222)
    at clickable-expander:68
    at clickable-expander:68
```

</details>

abiyasa:
Thanks a lot @mlrawlings I can confirm that this fixed on Marko 4.13.0

Issue #920, opened 2017-11-03
Uncaught TypeError: Cannot read property 'nextSibling' of null


bohdyone:
### Details

See an exception in the log. Also seeing occasional modelling errors -- such as some items in an each rendering twice. I'm guessing they're related. This happens when updates occur.

### Expected Behavior

No error in log and items appear only once even when re-rendering

### Actual Behavior

Items appear twice when updates trigger re-render

### Possible Fix

It may be a concurrency issue, as I'm using Marko to build a Chrome extension, and events can actually come in between any async processing.

<details><summary>Additional Info</summary>

### Your Environment

* Environment name and version (e.g. Chrome 39, node.js 5.4):
* Operating System and version (desktop or mobile):
* Link to your project:
Google Chrome Version 62.0.3202.75 (Official Build) (64-bit)
Mac OSX High Sierra

### Steps to Reproduce

Not sure how to reproduce it without using my project. I can send a link to it privately if needed.

### Stack Trace

```
view.js:4036 Uncaught TypeError: Cannot read property 'nextSibling' of null
    at Component.___forEachNode (view.js:4036)
    at Component.___detach (view.js:4027)
    at morphChildren (view.js:3236)
    at morphComponent (view.js:3118)
    at morphChildren (view.js:3242)
    at morphEl (view.js:3502)
    at morphChildren (view.js:3275)
    at morphEl (view.js:3502)
    at morphChildren (view.js:3275)
    at morphEl (view.js:3502)
___forEachNode @ view.js:4036
___detach @ view.js:4027
morphChildren @ view.js:3236
morphComponent @ view.js:3118
morphChildren @ view.js:3242
morphEl @ view.js:3502
morphChildren @ view.js:3275
morphEl @ view.js:3502
morphChildren @ view.js:3275
morphEl @ view.js:3502
morphChildren @ view.js:3275
morphComponent @ view.js:3118
morphChildren @ view.js:3242
morphdom @ view.js:3511
(anonymous) @ view.js:4011
batchUpdate @ view.js:2864
___rerender @ view.js:3994
update @ view.js:3949
updateComponents @ view.js:2845
updateUnbatchedComponents @ view.js:2817
(anonymous) @ view.js:2785
postMessage (async)
setImmediate @ view.js:2792
scheduleUpdates @ view.js:2836
queueComponentUpdate @ view.js:2902
___queueUpdate @ view.js:3926
___set @ view.js:2409
set @ view.js:2343
mouseOver @ view.js:23290
[...]

PR #1064, opened 2018-06-23
added failing test for issue #1059

cameronbraid:
I identified a bug in marko and managed to boil it down to a simple failing test case

## Checklist:

- [x ] My code follows the code style of this project.
- [ ] I have updated/added documentation affected by my changes.
- [x ] I have read the **CONTRIBUTING** document.
- [x ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.

PR #1066, opened 2018-06-24
Adding null check for nextSibling

BryceEWatson:
## Description

I've been debugging the below stack trace, and this is a simple fix that prevents it. I'm not 100% sure if there are other considerations here so please let me know, but this removes the exception when nextSibling is null.

```
Uncaught TypeError: Cannot read property 'nextSibling' of null
    at Component.___forEachNode (Component-91f50b8e.js:541)
    at getNodes (Component-91f50b8e.js:142)
    at Component.destroy (Component-91f50b8e.js:269)
    at detachedNodes.forEach (index-b9133752.js:809)
    at Array.forEach (<anonymous>)
    at morphdom (index-b9133752.js:801)
    at updateManager.___batchUpdate (Component-91f50b8e.js:515)
    at Object.batchUpdate [as ___batchUpdate] (update-manager-2ee5a85d.js:63)
    at Component.___rerender (Component-91f50b8e.js:498)
    at Component.update (Component-91f50b8e.js:451)
```
## Motivation and Context

I've seen this error reported in various contexts, but for me it is happening when I update state and remove a parameter from state. The removed parameter causes a UI element to be removed from the page, and it then proceeds to loop through the siblings of this component until nextSibling is null, at which point it will throw the exception. After the exception is thrown no further state updates are working.

A few issues I've seen this reported are https://github.com/marko-js/marko/issues/920 and https://github.com/marko-js/marko/issues/1051

This change fixes the above error by avoiding a null reference error, regardless of the root cause.

## Checklist:

- [NA] I have updated/added documentation affected by my changes.
- [NA] I have added tests to cover my changes.

### DIFF
--- a/src/components/Component.js
+++ b/src/components/Component.js
@@ -541,7 +541,7 @@ Component.prototype = componentProto = {
         for (;;) {
             var nextSibling = currentNode.nextSibling;
             callback(currentNode);
-            if (currentNode == endNode) {
+            if (currentNode == endNode || !currentNode.nextSibling) {
                 break;
             }
             currentNode = nextSibling;


cameronbraid:
I had done this myself for a while, however what its doing is treating the symptom not the cause and I suspect the PR will be rejected.

BryceEWatson:
Yes I agree it's just treating the symptom, but if there's no negative impact to doing so.. why not? I hope the cause is still fixed though, but this will at least let us continue working.

scttdavs:
If the next sibling being null is expected in some cases, then it may be ok
to do this and throw in a debug logging statement. But if not, then I think
the best thing it can do is fail early and loudly. Swallowing the error may
give it the appearance of it being fine, but there may be a fault somewhere
that is not discovered until later, and it will be impossible to debug
without any errors or stack traces.

BryceEWatson:
Agreed, this most likely isn't expected to be null and failing loudly is nice. If anything this PR can serve as a quick hack for anyone googling in the future.

Some additional context in case it helps identify the root cause:
It seems like the component being removed has `___endNode` assigned to a `<!--$marko-->` comment. It never runs across this comment and instead hits the end of the siblings.

Inspecting the html it looks like the comment is inserted at the bottom of the component, along with an extra closing `</div>` tag. This is where I hit a dead end, and someone with more marko knowledge would need to weigh in I think.

TLDR; Component.___endNode is incorrectly assigned to a marko comment, and is inserted oddly / incorrectly in the dom.

BryceEWatson:
Adding some observations:
Flow 1 - Error does occur during this flow:
1. Page loads with element visible on the page (state exists for it)
2. State update causes a re-render of the element, but it still exists.
3. State update causes the removal of the element, hits the original error.
Flow 2 - Error does *not* occur during this flow:
1. Page loads with element visible.
2. State update causes removal of element, this works fine.

Difference in call stacks:
Flow 1 (Errors):
See stack above, calls `Component.destroy` on all detached nodes.
Flow 2 (No error):
```
___forEachNode (Component-91f50b8e.js:541)
___detach (Component-91f50b8e.js:532)
morphChildren (index-7b0079e3.js:330)
insertVirtualNodeBefore (index-7b0079e3.js:79)
morphChildren (index-7b0079e3.js:441)
morphComponent (index-7b0079e3.js:156)
morphChildren (index-7b0079e3.js:347)
morphdom (index-7b0079e3.js:800)
updateManager.___batchUpdate (Component-91f50b8e.js:515)
batchUpdate (update-manager-2ee5a85d.js:63)
___rerender (Component-91f50b8e.js:498)
update (Component-91f50b8e.js:451)
updateComponents (update-manager-2ee5a85d.js:44)
updateUnbatchedComponents (update-manager-2ee5a85d.js:16)
(anonymous) (nextTick-browser-40824fd0.js:18)
postMessage (async)
setImmediate (nextTick-browser-40824fd0.js:27)
scheduleUpdates (update-manager-2ee5a85d.js:35)
queueComponentUpdate (update-manager-2ee5a85d.js:101)
___queueUpdate (Component-91f50b8e.js:417)
___set (State-legacy-623e5b28.js:91)
___replace (State-legacy-623e5b28.js:61)
replaceState (Component-91f50b8e.js:364)
```

cameronbraid:
@BryceEWatson FYI I have implemented a test case that produces this error https://github.com/marko-js/marko/pull/1064

BryceEWatson:
Great! Closing this PR since we dont understand the root cause + we have a test case now.

PR #1070, opened 2018-06-27
Fix client reordering when having multiple children

CestDiego:
## Description

I was looking at the client-reorder functionality and I'm sure that there is a bug in this code where it will continue appending the first element to the docFragment when there are modules with many children

## Motivation and Context

When using the await tag with code that renders multiple children we'll have problems reordering the content.

## Checklist:

- [X] My code follows the code style of this project.
- [NA] I have updated/added documentation affected by my changes.
- [in progress?] I have added tests to cover my changes.
- [X] All new and existing tests passed.

### DIFF
--- a/src/taglibs/async/client-reorder-runtime.js
+++ b/src/taglibs/async/client-reorder-runtime.js
@@ -24,7 +24,7 @@ function $af(
         len = childNodes.length;
 
         for (; i < len; i++) {
-            docFragment.appendChild(childNodes.item(0));
+            docFragment.appendChild(childNodes.item(i));
         }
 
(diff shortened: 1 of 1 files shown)


CestDiego:
I just ran the tests locally and this breaks quite a few tests :(

CestDiego:
I'll review how to fix them and add new tests as well. (All of the test
breaking have only one child)

CestDiego:
Found out that everytime we append to the documentFragment a child in the childNodes NodeList, the NodeList mutates and becomes smaller. Acting more like a queue instead of an array. 

This PR is unnecessary and I must have been doing something else wrong.

Issue #878, opened 2017-10-05
Support spread in attributes


mlrawlings:
### Description

Support the spread operator in tag attributes.
```
<tag ...attrs/>
```

### Why

We currently support two ways of passing dynamic attributes to a tag:

```marko
<custom-tag(input.attrs) data-foo=input.foo/>
<custom-tag ${input.attrs} data-foo=input.foo/>
```

Attributes should be the only way in which we pass data into tags (except for a _few_ exceptions like `if` and `for`).  The spread attribute is soon to be a JS standard and has a major benefit over the current solutions: order matters.

Using the parens as a tag argument (`<custom-tag(input.attrs)/>`) prevents you from ordering and even `<custom-tag ${input.attrs}/>` which could support ordering, doesn't.

Ordering gives you the ability to assign precedence, to create both default values and override values.

```marko
<custom-tag foo="default" ...input bar="override"/>
```

As a side note, the parens as an argument don't work for regular dom tags.
```marko
// no matter what `input.attrs` is, all you get is `<div data-foo>`
<div(input.attrs) data-foo=input.foo/>
```

Issue #656, opened 2017-04-06
Allow users of components to receive data in the body of a custom tag


mlrawlings:
### Description
_list.marko_
```marko
<ul>
    <li for(itemData in input.items)>
        <include(input.item, itemData)/>
    </li>
</ul>
```

_page.marko_
```marko
<list items=input.items>
    <@item(data)>
        ${input.name}
    </@item>
</list>
```

### Open Questions

Typically, `<tag(data)>` has been used to pass data to the tag, not receive data.  Is this a good pattern?

patrick-steele-idem:
I think it would be confusing to use the parens to define variables for the `renderBody()` function. I'll throw out something that might be a terrible idea:

```marko
<list items=input.items>
    <@item renderBody(data)>
        ${data.name}
    </@item>
</list>
```

I'm not really a fan of this either, because technically it becomes `function renderBody(out, data) {  }`


Maybe the following?:

```marko
<list items=input.items>
    <@item input(data)>
        ${data.name}
    </@item>
</list>
```

With destructuring?:

```marko
<list items=input.items>
    <@item input({name})>
        ${data.name}
    </@item>
</list>
```

patrick-steele-idem:
After giving it some more thought, it only makes sense to declare additional parameters for a `renderBody(out)` function if the the callee of the provided `renderBody` function is going to actually provide additional data as arguments (e.g. `input.renderBody(out, item)`). For that reason, I think it is the job for the UI component to force the user to declare the variable or possibly provide a default variable name (with the caveat that will be confusing to a user where the variable came from). For example. The `<list>` component might have the following `src/components/list/marko-tag.json`:

```json
{
  "<item>": {
    "var": { "nameFromAttribute": "var" }
  }
}
```

With this tag definition. The user would just need to do the following:

```marko
<list items=input.items>
    <@item var="theItem">
        ${theItem.name}
    </@item>
</list>
```

If the user didn't provide the `var` attribute then the user would get the following compilation error with the current implementation:

```
Attribute "var" is required
```

I think this is the best approach, but I think the following would be the next best alternative:

## `renderBody()` directive

By default, Marko generates adds a `renderBody(out)` method to the `input` object if a UI component has nested body content. If the user would like to control that function then they could use the `renderBody()` _directive_ (as an attribute on the tag):

```marko
<list items=input.items>
    <@item renderBody(theItem)>
        ${theItem.name}
    </@item>
</list>
```

The `renderBody()` directive could be documented as follows:

[...]

PR #1076, opened 2018-07-04
Tag params

mlrawlings:
## Description

Implements  #656/#851

Allow tag bodies to receive parameters

```marko
<mouse-position(mouse)>
  Mouse is at ${mouse.x}, ${mouse.y}.
</mouse-position>
```

Parameters can be destructured, etc.

```marko
<mouse-position({ x, y })>
  Mouse is at ${x}, ${y}.
</mouse-position>
```

**Feature Flag**

We used to use this syntax to pass arguments to the custom tag and although this is now deprecated and replaced by spread attributes (#878), in order to avoid a breaking change this feature will live behind a feature flag until Marko 5.

_`marko.json`_
```json
{
    "featureFlags": ["params"]
}
```

## Checklist:


src/compiler/util/parseJavaScriptParams.js
```diff
@@ -0,0 +1,13 @@
+"use strict";
+
+var ok = require("assert").ok;
+
+function parseJavaScriptParams(params, builder) {
+    ok(typeof params === "string", '"params" should be a string');
+    ok(builder, '"builder" is required');
+
+    var parsed = builder.parseExpression("function(" + params + "){}");
```
DylanPiercey: Would it be better to parse this as an arrow function? Then the line numbers would already be right if we use them eventually.

mlrawlings: Because `parseExpression` converts to Marko AST and we don't have an arrow function node, it returns a generic `Expression` node (which doesn't have the params attached).  So `function` it is.

DylanPiercey: 👍

cameronbraid:
Nice to see this work.. thanks.

Is there a way to globally defined a feature flag?

I don't currently rely on the old behaviour and I have many components that could make use of this feature but they don't currently have a tag def file for them

cameronbraid:
Sorry if this is off topic - could this same syntax be used for event delegation ?
mouse-position.marko :
```
class { 
  handleClick() {}
}
<${input} x=100 y=200 onClick=delegateTo('handleClick')/>
```

```
<mouse-position({ x, y, onClick })>
  Mouse is at ${x}, ${y}.
<div on-click(onClick)>click me</div>
</mouse-position>
```

would cause the transcluded on-click handler to delegate to the handler that mouse-position provides as 'handleClick'

src/compiler/util/parseJavaScriptParams.js
mlrawlings: So I was concerned with some of the changes I made to `parseJavaScript.js`.  By using `esprima` directly here I don't need to make those changes and can also use an arrow function.

Issue #947, opened 2017-12-13
After redraw input component is regenerated with incorrect key. 


matisskeiris:
## Bug Report
After redraw marko regenarates key for component. 

### Details
I have a modal component, that can be shown and hidden. 
```
<div class=['modal large-modal', !state.visible && 'hidden' ]>
    <div class="modal-background" onClick("handleMaskClick") />
    <div class="exit-button" />
    <div class="modal-container">
        <div class="modal-body" key="body">
            <include (input.renderBody) />
        </div>
        <div class="modal-footer">
            <include (input.actions) />
        </div>
    </div>
</div>
```
I am using this component to display a timer component
```
<large-modal key="popup-window" on-hide('pauseTimer', 'timer')>
    <@renderBody> 
        <timer key="timer" />
    </@renderBody>
</large-modal>
```
This timer component has a method pause, that would pause the timer after modal is closed. To do that, i have added method on hide event on modal, that looks like this. 
```
pauseTimer(timerKey) {
    this.getComponent(timerKey).pause();
}
```
This functionality does not work, as after each manipulation on modal, that is state.visibility changes, the contents are redrawn, including timer key="timer" and marko sets this components id now to timer_1, which makes my functionality with call pauseTimer('timer') to not work, as this component no longer is identified by this key. 

### Expected Behavior
I am expecting key to not change in scenario when component is redrawn.

DylanPiercey:
@matisskeiris I created a [branch](https://github.com/marko-js/marko/tree/issue-947) which adds a [test](https://github.com/marko-js/marko/commit/b566e8fddc91b4dbc646ce9afeb7b662dc140009) for this but I was unable to reproduce the issue.

Would you be able to pull down this branch and help in creating a failing test?

You can run the test added in my branch by running:
```console
node ./test/__runner__/browser/cli.js --page preserve-transcluded-key-on-state-change --automated
```

jasonmacdonald:
@DylanPiercey We actually hit this same problem today in a drop-down component we built where the body of the dropdown is a transcluded component. If the conditional is on the include inside another component, the key gets incremented the second time the component is rendered, but the parent context still tries to lookup the key using the original key name with no index (because the parent is not aware that the component was destroyed and recreated. This seems to happen because when the component is rerendered, it uses the parent's key sequence variable which has the key (despite the child being destroyed), and since the destruction of the transcluded component happened in a child component, the parent component is not rerendered so the key sequence is not rebuilt.

[...]

PR #1086, opened 2018-07-14
fixes #1022 - transcluded keys don't match, stateful component not preserved

### LINKED ISSUE #1022: forceUpdate/include semantics have changed since 4.5.0-beta.2
I am having an issue in 4.7.5 (and other versions, see below) with calling forceUpdate causing state within included components to be lost.  The components are being destroyed and re-created.

I have setup a standalone project using yarn/wepack dev server/browser-refresh

https://github.com/cameronbraid/marko-forceUpdate-include-issue

To exercise the issue, click both the counting buttons, then click the forceUpdate buttons.  The forceUpdate buttons should not reset any of the counting buttons to zero.

To run :

* `yarn`  to install deps
* `yarn dev` in one shell - for webpack dev server
* `yarn server` in another shell - for browser-refresh server

It runs server at http://localhost:8080 

I have setup the project with marko@4.5.0-beta.2 since this demonstrates what I believe is correct semantics for forceUpdate - that component state is retained when forceUpdate is called.

It appears that 4.5.0-beta.3 introduces the problem

* `yarn add marko@4.5.0-beta.3` 
* `yarn add marko@4.7.5`
* `yarn add marko@4.9.2`

mlrawlings:
## Description

Fixes Issue #1022: Because keys can be repeated in Marko, a component has a KeySequence that keeps track of the number of times a key has been rendered.  This gets reset each time the component is re-rendered, but in the case of transcluded content, the owner component (the component in which a node is defined) is not the only cause of re-renders.  The immediate parent can cause a rerender and in this case, the owner component's KeySequence picks up where it left off, generating incorrect keys.

In practice, this means nodes get thrown away that shouldn't be.  And this is especially problematic when the node that gets thrown away is a stateful component as its internal state gets lost.

This PR fixes this issue by using the immediate parent's KeySequence instead of the owner component's.  Keys are still scoped to the owner component, so `getComponent` still works.

## Checklist:

### DIFF
--- a/test/components-browser/fixtures/transclusion-rerender-stateful/test.js
+++ b/test/components-browser/fixtures/transclusion-rerender-stateful/test.js
@@ -5,17 +5,15 @@ module.exports = function(helpers) {
     var container = root.getComponent("container");
     var counter = root.getComponent("counter");
 
-    expect(document.body.textContent).to.equal("0");
+    expect(helpers.targetEl.textContent).to.equal("0");
 
     counter.increment();
     counter.update();
 
-    expect(document.body.textContent).to.equal("1");
+    expect(helpers.targetEl.textContent).to.equal("1");
 
     container.forceUpdate();
     container.update();
 
-    expect(document.body.textContent).to.equal("1");
+    expect(helpers.targetEl.textContent).to.equal("1");
 };
-
-module.exports.fails = "issue #1022";
--- a/src/components/renderer.js
+++ b/src/components/renderer.js
@@ -22,7 +22,7 @@ function resolveComponentKey(
         return key.substring(1);
     } else {
         parentComponentDef = parentComponentDef || ownerComponentDef;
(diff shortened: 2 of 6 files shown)


test/components-browser/fixtures/transclusion-rerender-stateful/test.js
```diff
@@ -5,17 +5,15 @@ module.exports = function(helpers) {
     var container = root.getComponent("container");
     var counter = root.getComponent("counter");
 
-    expect(document.body.textContent).to.equal("0");
+    expect(helpers.targetEl.textContent).to.equal("0");
```
mlrawlings: When hydrating, the body contains the hydration code as well as the `0`, so I've switched to `targetEl` instead.

mlrawlings:
@jasonmacdonald would you be willing to test if this fixes your issue in #947?

PR #1131, opened 2018-09-16
Fixed broken links

thordy:
The links to server-side framework integrations all return 404, so fixed the links to point to the correct pages

## Description

Links were incorrect, giving a 404, so resolved them to the proper pages

## Checklist:

- [ x ] I have read the **CONTRIBUTING** document and have signed (or will sign) the CLA.
- [ x ] I have updated/added documentation affected by my changes.
- [ x ] I have added tests to cover my changes.

### DIFF
--- a/docs/server-side-rendering.md
+++ b/docs/server-side-rendering.md
@@ -24,10 +24,10 @@ module.exports = function(req) {
 
 > **ProTip:** Marko also provides server-side framework integrations:
 >
-> - [express](./express.md)
-> - [hapi](./hapi.md)
-> - [koa](./koa.md)
-> - [huncwot](./huncwot.md)
+> - [express](./express)
+> - [hapi](./hapi)
+> - [koa](./koa)
+> - [huncwot](./huncwot)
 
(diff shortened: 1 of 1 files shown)


austinkelleher:
Hey @thordy! Thanks for taking a stab at this. I think this isn't quite right. We want the links in the markdown files to include the extensions so that they will properly redirect in the GitHub UI. Currently how we tackle this is by removing the `.md` from file extensions and generating the correct path during build time of the markojs.com website. There is probably a bug in this code here, but I haven't gotten a chance to debug it. Here's the code if you're interested: https://github.com/marko-js/markojs-website/blob/master/routes/docs/util/markdown.js#L86

thordy:
Hi @austinkelleher , Sorry about that, I definitely misunderstood how that works. I took a stab at fixing it, and it seems to me it's just an issue with the regex in the replace. I had some issues completely verifying the change, but running it locally, it will now resolve those links correctly as expected. 

I created a pull request to fix the issue.
https://github.com/marko-js/markojs-website/pull/37

Issue #1134, opened 2018-09-27
Server rendered content removed from looped @tag


DylanPiercey:
### Details

Currently if you have preserved content that is ultimately passed to a repeated (in a loop) `@tag` the server rendered content will be discarded on mount.

### Expected Behavior

The content should be preserved.

### Actual Behavior

It renders properly server side and is removed when components initialize in the browser.

<details><summary>Additional Info</summary>

### Your Environment

- Environment name and version (e.g. Chrome 39, node.js 5.4): Node 10
- Operating System and version (desktop or mobile): MacOS

### Steps to Reproduce

Failing test case PR: #1135

</details>

PR #1135, opened 2018-09-27
Fix for preserved content in repeated attribute tags.

DylanPiercey:
## Description

Adds a test case and potential fix for issue #1134.

This fix is to include the owner key in loop keyscopes. It seems the main issue is that the attribute tag was being keyed relative to the parent not the owner and the keyscope var (which was in the owner) was colliding with one in the parent.

## Checklist:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
     "name": "marko",
-    "version": "4.13.6-0",
+    "version": "4.13.6-1",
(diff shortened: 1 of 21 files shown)


DylanPiercey:
Updated this PR to include a potential fix. I feel that there is probably a better way to fix this though as the generated keys will be much longer.

mlrawlings:
No.

PR #1158, opened 2018-10-26
[MARKO-3] - <assign> and <var> tag migration

timlauter:
Added `<var>` and `<assign>` migration transforms so users can still use the deprecated syntax, but allowing the reference to not be a breaking change in Marko V4.

## Description

Both the `<assign>` and `<var>` tag have their attributes looped through to get transformed from a type of `HtmlAttribute` to a `Scriptlet`.

In the assign tag, one of two cases happen. Either the attribute name gets assigned a value, or the attribute name's value remains `undefined`. During this assignment, the new `Scriptlet` that is created is inserted as a sibling before the original elNode. After, the original elNode is detached.

The var tag has similar functionality to the assign tag. However, there are two differences. While looping through the attributes of the elNode, the attribute names are actually being defined as a variable and assigned rather than only being referenced. Again, during this assignment, the new `Scriptlet` that is created is inserted as a sibling before the original elNode. Before the detachment of the original elNode, it's children are looped over and inserted as a sibling before the original elNode to not lose the children that are referencing the declared variable.

Allows users to reference deprecated code without it breaking on compilation because the way the user wrote to code is transformed into how the code should be written.

## Checklist:


DylanPiercey:
@timlauter this looks good 😄! Would you mind updating the two failing snapshot tests to get the CI passing?

PR #1166, opened 2018-11-08
Add support for file name to contain [some word]

aganeshalingam:
Add support for file name to contain [mobile]


seangates:
That's a lot of extra whitespace changes.

seangates:
@aganeshalingam 👍

DylanPiercey:
Let it be know that I think this is a really strange PR and that this intent is to better support for https://github.com/eBay/arc.

Having said that the impact here is really small. Will merge this, but hopefully we find a different solution so that Marko does not need to know about arc.

PR #1177, opened 2018-11-27
Init components on ready

scttdavs:
## Description

Marko includes an inline script tag to initialize all components on the page. It must be added after all the component resources are loaded, and it does this by inserting the inline script just before the closing body tag, insuring all js for Marko components would be loaded already.

However, a common js optimization is to use deferred scripts. But since inline scripts cannot be deferred, they must be wrapped in a document ready listener to ensure the js has been loaded.

This wraps the component initialization code in a document ready listener.

## Checklist:


scttdavs:
nvm, as it was designed, this should work even when other resources are deferred.

PR #1180, opened 2018-12-03
Migration stage

DylanPiercey:
## Description

Adds a migration stage to the compiler which is exposed through the taglib api via a new `migrate` option on both tags and taglibs.

This PR also moves over some of the existing transform based migrations (assign, var, invoke, and imperative rendering). It also adds a new migration for (and deprecates) inline control flow directives (<div if(x)>`).

This piggybacks off of #1179.

## Checklist:


ianvonholt:
Was there any particular reason to remove inline control statements?

`<div if(x)>Blargh</div>` 

is a bit more streamline than 

```
<if(x)>
  <div>Blargh</div>
</if>
```

DylanPiercey:
@ianvonholt it was something we have been talking about for a while, although I wish we were a bit more transparent about it.

Here is the reason:

We have seen this particular syntax is frequently used in ways that have caused confusion among developers. Specifically it's very easy to glance over where these (important) control flow statements are taking place. I think [our example](https://github.com/marko-js/marko/wiki/Deprecation%3A-control-flow-directive) illustrates one of the reasons behind this, which is that it can simply be very tricky to keep on top of formatting this in a readable way especially when attributes start to pile up.

The value add for this feature was primarily around terseness and readability (which is not something we want to move away from) however I think In this case it ended up becoming more of a foot gun in the language. In examples (like the one you showed) it does look nice, but when there are other attributes on the tag (as is often the case in real-life, non-example code) readability actually suffers.

On top of this, these directives are yet another thing that new developers have to learn when on boarding to the language and also a feature that makes the framework a bit harder to maintain (two ways to do everything can really complicate the code base).

Because of this, and because it is an extremely easy feature to automatically migrate, we've decided  deprecate it here in Marko 4. It will be removed in Marko 5.

Our recent work has been focused on creating a migration layer in the Marko compiler so that most of these features you are seeing deprecation warnings for can be automatically migrated by [marko migrate](https://github.com/marko-js/cli/blob/master/packages/migrate/README.md).

mlrawlings:
As @DylanPiercey mentioned, this is something we've been discussing for a while.  It happened here because not doing so would have required some extra effort in implementing some of the things in this PR (and future migration related PRs).

One of the reasons we do deprecations in advance is so that users can see these warnings and take things up with us if we're missing a use case.  So this isn't 100% set in stone, but I do think it's the right path forward.

ianvonholt:
@mlrawlings @DylanPiercey Thank you for the input, I do get the reasons for deprecation when there are elements that have a lot of attributes. 

Speaking of Marko v5, is there a roadmap of features that are being considered?

Issue #1181, opened 2018-12-06
Marko 4.14.0 is stripping arrow functions sometimes


cameronbraid:
Repro https://github.com/cameronbraid/marko-arrow-function-issue

### Marko Version: 4.14.0

Broken.marko
```
static {
    function runIt(fn) {
        fn()
    }
}
class {
    onCreate() {
        this.state = {
            message : "didnt run it"
        }
        if (true) {
            runIt(()=>{
                this.state.message = "ran it"
            })
        }

    }
}
<div>Broken [${state.message}]</div>
```

Works.marko
```
static {
    function runIt(fn) {
        fn()
    }
}
class {
    onCreate() {
        this.state = {
            message : "didnt run it"        
        }
        runIt(()=>{
            this.state.message = "ran it"
        })
    }
}
<div>Works [${state.message}]</div>
```

Broken.marko.js (see  ```if (true) {}```)
```
// Compiled using marko@4.14.0 - DO NOT EDIT
"use strict";

var marko_template = module.exports = require("marko/src/html").t(__filename),
    marko_component = {
        onCreate: function() {
          this.state = {
              message: "didnt run it"
            };

          if (true) {}
        }
      },
    marko_componentType = "/marko-dom-issue-no-mobx$0.0.1/src/components/Broken.marko",
    components_helpers = require("marko/src/components/helpers"),
    marko_renderer = components_helpers.r,
    marko_defineComponent = components_helpers.c,
    marko_helpers = require("marko/src/runtime/html/helpers"),
    marko_escapeXml = marko_helpers.x;

function runIt(fn) {
    fn()
};

function render(input, out, __component, component, state) {
  var data = input;

  out.w("<div>Broken [" +
    marko_escapeXml(state.message) +
    "]</div>");
}

marko_template._ = marko_renderer(render, {
    ___type: marko_componentType
  }, marko_component);

marko_template.Component = marko_defineComponent(marko_component, marko_template._);

marko_template.meta = {
    id: "/marko-dom-issue-no-mobx$0.0.1/src/components/Broken.marko",
    component: "./Broken.marko"
  };
```

DylanPiercey:
The patch was released in 4.14.1, thanks for all of the help with this one 😄.

PR #1182, opened 2018-12-06
added failing test case for #1181

cameronbraid:
added failing test case for https://github.com/marko-js/marko/issues/1181


DylanPiercey:
Thanks for this, very helpful!

Just want to add that I've taken a look at this as well and the issue is actually that I messed up when adding the new nodes (IfStatement, ForStatement and While) to `parseJavascript`. Every other converter returns null immediately if any nested conversions return null, however I was just letting it pass through.

I'm going to merge this PR and put up a quick patch PR.

PR #1183, opened 2018-12-06
support arrow functions, with test cases

cameronbraid:
Implement codegen support for arrow functions


cameronbraid:
i'll add support for the rest of the ArrowFunctionExpression syntax tree and re-submit

PR #1185, opened 2018-12-06
Added migrator for async-fragment & bugfix for Normalizer.js

BhavinPatel04:
## Description
I have made the below changes to deprecate the `<async-fragment>`:

- Added new migrator file to migrate:
    - `<async-fragment>` to `<await>`
    - `<async-fragments>` to `<await-reorder>`
    - `<async-fragment-placeholder>`  to `<await-placeholder>`
    - `<async-fragment-timeout>`  to `<await-timeout>`
    - `<async-fragment-error>`  to `<await-error>`
- Added test cases for above
- Removed the transformer for `<async-fragment>`
- Bugfix for Normalizer

The above changes were required to replace `<async-fragment>` with `<await>`

## Checklist:


src/taglibs/migrate/async-fragment-tag.js
```diff
@@ -0,0 +1,59 @@
...
+    "async-fragment": "await",
+    "async-fragments": "await-reorderer",
+    "async-fragment-placeholder": "await-placeholder",
+    "async-fragment-timeout": "await-timeout",
+    "async-fragment-error": "await-error"
+};
+module.exports = function migrator(oldNode, context) {
+    const oldTag = oldNode.tagName;
+    const newTag = newTags[oldTag];
+    var provider;
```
DylanPiercey: Minor nitpick, would you mind changing `var` to `let` here?

src/taglibs/migrate/async-fragment-tag.js
```diff
@@ -0,0 +1,59 @@
...
+    var varName;
+    var argument;
+    const attributes = oldNode.attributes;
+    const builder = context.builder;
+    migrateControlFlowDirectives(oldNode, context);
```
mlrawlings: The var/assign/invoke tags needed to do this because they were being migrated to a scriptlet which is not an HtmlElement and doesn't have attributes, so the control flow would have been lost if it wasn't handled here.

```js
migrateControlFlowDirectives(oldNode, context);
oldNode.setTransformerApplied(migrateControlFlowDirectives);
```

However, I think these lines aren't necessary in this case because the attributes get copied over to the new element if the migration hasn't already run.

DylanPiercey:
@BhavinPatel04 could you update the snapshot for the `taglib-lookup/fixtures > forEachTag` test?

src/taglibs/migrate/async-fragment-tag.js
```diff
@@ -0,0 +1,59 @@
...
+    oldNode.setTransformerApplied(migrateControlFlowDirectives);
+
+    context.deprecate(
+        `The "<${oldTag}>" tag is deprecated. Please use "<${newTag}>" instead. See: https://github.com/marko-js/marko/wiki/Deprecation:-var-assign-invoke-tags`
```
DylanPiercey: We will need to update this error message and create a new wiki page here: https://github.com/marko-js/marko/wiki/Deprecation:-async-fragment

This should follow a similar format to https://github.com/marko-js/marko/wiki/Deprecation:-var-assign-invoke-tags

DylanPiercey:
Thanks for this @BhavinPatel04! Looks good to me now. 😄.

PR #1190, opened 2018-12-11
Added migrators for w-* attrs & code restructure

BhavinPatel04:
## Description

Added migrators for 
`w-id`  →  `key`
`w-for` & `for-key`  →  `for:scoped`
`:key`  →  `:scoped`
`w-preserve`  →  `no-update`
`w-preserve-attrs`  →  `:no-update`

Restructured the code in taglibs/migrate

## Checklist:


src/taglibs/migrate/common/event-handlers/w-preserve-attrs.js
```diff
@@ -0,0 +1,23 @@
...
+
+            const values = attr.value.toString();
+            context.deprecate(
+                `The "w-preserve-attrs" attribute is deprecated. Please use ":no-update" modifier instead. See: https://github.com/marko-js/marko/wiki/Deprecation:-w-*-Atrributes`
+            );
+
+            values.split(",").forEach(val => {
+                const attrValue = el.getAttributeValue(val);
+                if (attrValue)
+                    el.setAttributeValue(`${val}:no-update`, attrValue);
```
DylanPiercey: In the `else` here we should still add `${val}:no-update` with an empty attribute since the users intent may be to have the attribute be unset during the initial render and then mutated with JS later.

src/taglibs/migrate/common/event-handlers/w-preserve.js
```diff
@@ -0,0 +1,21 @@
...
+        if (!name || !Object.keys(newTags).includes(name)) {
+            return;
+        }
+
+        const newAttrName = newTags[name];
+        context.deprecate(
+            `The "${name}" attribute is deprecated. Please use "${newAttrName}" attribute instead. See: https://github.com/marko-js/marko/wiki/Deprecation:-w-*-Atrributes`
+        );
+
+        attr.name = newAttrName;
```
DylanPiercey: I could be wrong but I think the `if` variants have their attribute values as a string and we will need to parse it as javascript node.

src/taglibs/migrate/util/addIdScopedAttr.js
```diff
@@ -0,0 +1,16 @@
+module.exports = function addIdScopedAttr(context, el, attrValue) {
+    const templateRoot = el.parentNode;
+    const walker = context.createWalker({
+        enter(node) {
+            if (node.type == "HtmlElement" && node.hasAttribute) {
```
DylanPiercey: Looks good but I think that we also need to check for `type === "CustomTag"`. It would be simplest to only check for `node.hasAttribute` as only `HTMLElement`'s and `CustomTags`'s should have that method.

src/taglibs/migrate/util/addIdScopedAttr.js
```diff
@@ -0,0 +1,16 @@
+module.exports = function addIdScopedAttr(context, el, attrValue) {
+    const templateRoot = el.parentNode;
```
DylanPiercey: `el.parentNode` is not the actual template root. We actually want to use `context.root`.

src/taglibs/migrate/common-tag-migrator.js
```diff
@@ -1,8 +1,12 @@
 const fs = require("fs");
 const path = require("path");
-const commonMigrators = fs
-    .readdirSync(path.join(__dirname, "./common/"))
-    .map(dir => require(`./common/${dir}`));
+const getMigrators = d =>
```
DylanPiercey: I think we should revert this such that the `common` migrations is a flat structure (move all the `w-` stuff up a level).

As it stands right now all of these are in an `event-handlers` folder which is not totally accurate and I think unnecessary.

BhavinPatel04: Cool. Let me structure it.

BhavinPatel04:
Glad to be of help! :D

PR #1199, opened 2018-12-18
Migrate w-body attr in favor of using dynamic tags

BhavinPatel04:
## Description

Replacing w-body attribute in favor of using <$dynamicTag />
Added test cases for the migration

## Checklist:


src/taglibs/migrate/all-tags/w-body.js
```diff
@@ -0,0 +1,35 @@
+const printJS = require("../util/printJS");
+
+module.exports = function migrate(el, context) {
+    if (el.hasAttribute("w-body")) {
+        context.deprecate(
+            `The "w-body" attribute is deprecated. Please use "<\${dynamicTag}/>" attribute instead. See: https://github.com/marko-js/marko/wiki/Deprecation:-w-*-Atrributes`
```
DylanPiercey: This message should be:

```
`The "w-body" attribute is deprecated. Please use the "<${dynamicTag}/>" tag instead. See: https://github.com/marko-js/marko/wiki/Deprecation:-w-*-Atrributes`
```

src/taglibs/migrate/all-tags/w-body.js
```diff
@@ -0,0 +1,35 @@
...
+        );
+
+        const builder = context.builder;
+        const bodyValue =
+            el.getAttributeValue("w-body") || builder.identifier("input");
+        el.removeAttribute("w-body");
+
+        const ifExpressionArg = `typeof ${builder.identifier(
```
DylanPiercey: Here we should just be checking the `printJS` of the `bodyValue` like so:

```js
const ifExpressionArg = `typeof ${printJS(bodyValue, context)}`;
```

test/migrate/fixtures/w-body-attr-val/snapshot-expected.marko
```diff
@@ -0,0 +1,8 @@
+<div>
+  <if(typeof input.renderBody === "string")>
```
DylanPiercey: The change recommended above should fix this `<if(` which should be checking `typeof input.test.renderBody`.

src/taglibs/migrate/marko.json
```diff
@@ -99,5 +99,11 @@
         "preserve-name": true,
         "autocomplete": [],
         "deprecated": true
+    },
+    "@w-body": {
+        "preserve-name": true,
+        "autocomplete": [],
+        "deprecated": true,
+        "set-context-flag": "hasLegacyWidgetAttr"
```
DylanPiercey: I think we can remove this `set-context-flag` option. @mlrawlings?

PR #1200, opened 2018-12-20
Migrate legacy dynamic attributes

BhavinPatel04:
## Description
Added a migrator to deprecate `<div ${attributes} />` in favor of using `<div ...attributes />`
Added test cases for the above

## Checklist:


src/taglibs/migrate/all-tags/dynamic-attributes.js
```diff
@@ -0,0 +1,19 @@
+module.exports = function migrate(el, context) {
+    el.forEachAttribute(attr => {
+        const name = attr.name;
+        const value = attr.value;
+        if (
+            name == undefined &&
```
DylanPiercey: This should use the same check that is used [here](https://github.com/marko-js/marko/blob/7d37fe36347fc75deb4148e9196f57e51b13121c/src/taglibs/core/core-transformer.js#L134) (`!attr.name && !attr.spread`). The implementation here will fail if you have something like `...({ a : 1 })` or any other type of node that isn't and identifier or member expression.

Also the previous warning should be removed from [here](https://github.com/marko-js/marko/blob/7d37fe36347fc75deb4148e9196f57e51b13121c/src/taglibs/core/core-transformer.js#L134) as well.

src/taglibs/migrate/all-tags/dynamic-attributes.js
```diff
@@ -0,0 +1,19 @@
...
+            (value.type == "MemberExpression" || value.type == "Identifier")
+        ) {
+            context.deprecate(
+                'The "${attributes}" is deprecated. Please use "...attributes" modifier instead. See: https://github.com/marko-js/marko/wiki/Deprecation:-w-*-Atrributes'
+            );
+            attr.spread = true;
```
DylanPiercey: This line can be removed if we switch to removing the attribute and adding the spread attribute to the end.

src/taglibs/migrate/all-tags/dynamic-attributes.js
```diff
@@ -0,0 +1,19 @@
...
+            if (value.property && value.property.value == "*") {
```
DylanPiercey: This check should not be needed, to should be safe to just remove the existing dynamic attribute and add a new spread attribute like so:

```js
el.removeAttribute(attrName);
el.addAttribute({ spread: true, value: value });
```

test/render/fixtures/attrs/expected.html
```diff
@@ -1 +1 @@
-<div data-encoding="&quot;hello&quot;" style="background-color: #FF0000; &lt;test>" class="my-div" checked>Hello World!</div>
```
DylanPiercey: This test should not need to change, doing so would be a breaking change. If the order of the dynamic tags moves to the end this should be fixed.

test/render/fixtures/script-nonce/expected.html
```diff
@@ -1,3 +1,3 @@
-<script foo="bar" nonce="foo" hello="world">
```
DylanPiercey: This test should not need to be changed.

BhavinPatel04:
Closing this since a new invalid user `Patel` is added to committers.

PR #1204, opened 2018-12-26
Migrate include, layout-use and layout-placeholder tags

DylanPiercey:
## Description

Adds migrator for the `<include>` tag to become either a `<${dynamic}/>` tag or a `${placeholder}`.

Because both `layout-use` and `layout-placeholder` previously transformed into include tags, similar migrations were added for these.

## Checklist:


src/runtime/helpers.js
```diff
@@ -157,6 +158,8 @@ var helpers = {
                         r[removeDashes(key)] = attrs[key];
                         return r;
                     }, {});
+                } else if (attrs == null) {
+                    attrs = {};
```
DylanPiercey: This was done because previously it was possible to have undefined input passed to a renderer with the dynamic tag. I believe this was an error.

test/inline/fixtures/layout-template/test.js
```diff
@@ -0,0 +1 @@
+module.exports.fails = "TODO: fix or remove";
```
DylanPiercey: This single test fails because when using `inline templates` there is no component set on the componentContext? @mlrawlings would you mind taking a look at this?

PR #1206, opened 2018-12-26
Added a migrator for <layout-put>

BhavinPatel04:
## Description
Added a migartor to migarte `<layout-put>` in favor of using `<@tags>`

## Checklist:


src/taglibs/migrate/layout-put-tag.js
```diff
@@ -0,0 +1,35 @@
+const commonTagMigrator = require("./all-tags");
+
+module.exports = function migrator(oldNode, context) {
+    const attributes = oldNode.attributes;
+    commonTagMigrator(oldNode, context);
+    oldNode.setTransformerApplied(commonTagMigrator);
+
+    context.deprecate(
+        'The "<layout-put>" tag is deprecated. Please use "<@tags>" instead. See: https://github.com/marko-js/marko/wiki/Deprecation:-var-assign-invoke-tags'
```
DylanPiercey: We'll need a wiki page for this. I'm thinking it'd make sense to just have a shared page for all `layout` tags.

BhavinPatel04: Cool. I will create a wiki page for `layout` and update it here

src/taglibs/migrate/layout-put-tag.js
```diff
@@ -0,0 +1,35 @@
...
+        return oldNode;
+    }
+
+    const name = oldNode.getAttributeValue("into").value;
+    const value = oldNode.getAttributeValue("value");
+
+    oldNode.removeAttribute("into");
+    oldNode.removeAttribute("value");
+
+    const newNode = context.createNodeForEl(
```
DylanPiercey: We shouldn't be using `createNodeForEl` here (should only happen in transformers). Instead we should use `builder.htmlElement` as we have been.

src/taglibs/migrate/layout-put-tag.js
```diff
@@ -0,0 +1,33 @@
...
+    }
+
+    const builder = context.builder;
+    const name = oldNode.getAttributeValue("into").value;
+    const value = oldNode.getAttributeValue("value");
+
+    oldNode.removeAttribute("into");
+    oldNode.removeAttribute("value");
+
+    const newNode = builder.htmlElement("@" + name, oldNode.getAttributes());
```
DylanPiercey: Minor nitpicks:

Let's reuse `attributes` from above instead of using `getAttributes`.

src/taglibs/migrate/layout-put-tag.js
```diff
@@ -0,0 +1,33 @@
...
+
+    if (value) {
+        newNode.appendChild(context.builder.text(value));
```
DylanPiercey: Let's reuse the `builder` var from above instead of `context.builder` here.

PR #1210, opened 2018-12-27
Migrate unless tag

BhavinPatel04:
## Description

Added a migrator for `unless` tag
`unless` is deprecated in favor of using `if` tag

## Checklist:


src/taglibs/migrate/unless-tag.js
```diff
@@ -0,0 +1,39 @@
+const printJS = require("./util/printJS");
+const commonTagMigrator = require("./all-tags");
+
+module.exports = function migrator(oldNode, context) {
+    if (oldNode.tagName !== "unless") {
```
DylanPiercey: We should not need this check since the migrator only runs on `<unless>` tags.

src/taglibs/migrate/unless-tag.js
```diff
@@ -0,0 +1,39 @@
...
+        context.addError("Invalid <unless> tag. Attributes not allowed.");
+        return oldNode;
+    }
+
+    const builder = context.builder;
+    const newNode = builder.htmlElement(
+        "if",
+        undefined,
+        [],
+        printJS(`!${argument}`, context),
```
DylanPiercey: This line should switch to use `builder.negate` like so:

```js
printJS(builder.negate(argument), context),
``` 

This will fix cases like `<unless(x && y)>` which would currently become `<if(!x && y)>` (we should add a test case for that as well).

test/migrate/fixtures/unless-tag/tempalte.marko
```diff
@@ -0,0 +1,12 @@
+<unless(x)>
```
DylanPiercey: This filename has a typo (should be template.marko).

test/migrate/fixtures/unless-tag/tempalte.marko
```diff
@@ -0,0 +1,12 @@
...
+    <div />
+</unless>
+<unless(true)>
+    <div />
+</unless>
+<unless(true sdsds)>
```
DylanPiercey: We probably don't need to test syntax errors here since they are handled in the builders. Also the syntax error here is going to cause this migrator to never output. If we did wan't to test syntax errors they would have to be in a separate test.

PR #1217, opened 2019-01-03
w-id migration key only for components

mlrawlings:
## Description

This PR updates the `w-id` migration to omit `id:scoped` and only output `key` for component tags.

HTML and `${dynamic}` tags still receive both `key` and `id:scoped`.   This is consistent with the behavior from Marko 3 as HTML tags had an `id` attribute set by `w-id` and dynamic tags in Marko 3 could only render html elements.  Components, however, did not receive an `id` key in their input props.

Additionally, without this change, a custom tag that has defined attributes in a `marko-tag.json` will throw if `w-id` is used on it because the `id:scoped` attribute gets added and that is not a valid attribute for the tag.

## Checklist:

- [ ] ~I have updated/added documentation affected by my changes.~


src/taglibs/migrate/all-tags/w-id.js
```diff
@@ -9,6 +9,9 @@ module.exports = function migrate(el, context) {
     );
 
     el.setAttributeValue("key", attr.value);
-    if (!el.hasAttribute("id")) el.setAttributeValue("id:scoped", attr.value);
+    const isHTML = el.tagDef && el.tagDef.html;
+    const isDynamic = /^\$\{.+\}$/.test(el.tagName);
```
DylanPiercey: I think it would be better to check for `rawTagNameExpression` or `tagNameExpressions` instead of the RegExp like so `const isDynamic = Boolean(el.rawTagNameExpression)`.

Otherwise the PR LGTM.

PR #1229, opened 2019-01-11
Add migrator for legacy nested tag syntax

DylanPiercey:
## Description

Previously the syntax for nested tags / layout was `<tag-name:nested-name>`. This has been replaced with `<@nested-named>`. This PR adds a deprecation warning and migrated the old syntax.

## Checklist:


src/taglibs/migrate/all-tags/legacy-nested-tag.js
```diff
@@ -0,0 +1,16 @@
+const legacyNestedTagSyntax = /^.*:(.*)$/;
```
austinkelleher: I doubt this is a problem, but this will technically match a tag in this format: `<:helloworld>`

PR #1264, opened 2019-02-20
provide hook to get serialized globals before component init

cameronbraid:
this PR exposes a new method require("marko/components").initGlobals() which allows access to the serialized globals before a call to require("marko/components").init()

I need this hook so that I can do some initialization client side, using my server side state, before marko builds the client side component tree. Part of which may be fetching a lazy loaded webpack module that is required to render the components

### DIFF
--- a/src/components/index-browser.js
+++ b/src/components/index-browser.js
@@ -7,6 +7,8 @@ require("./ComponentsContext").___initClientRendered =
 
 exports.getComponentForEl = componentsUtil.___getComponentForEl;
 exports.init = window.$initComponents = initComponents.___initServerRendered;
+exports.initGlobals = window.$initGlobals =
+    initComponents.___initServerRenderedGlobals;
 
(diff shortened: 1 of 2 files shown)


cameronbraid:
I'm happy to write a test case for this if you can point me to the best place to put it

casebuiltfurniture:
Ok

DylanPiercey:
@cameronbraid sorry for taking so long to respond to this PR. So the reason we've been hesitant to merge in this one is because for the most part we don’t really consider what is serialized in the output to be a part of our public API. This output can change due to small internal changes and we'd like to reserve the ability to make improvements here without creating breaking changes to our public API.

I think things like lazy loading modules and anything else I can think of that would require this should probably have a more official solve (eg we want to make the `<await>` tag work in the browser) and I think we can solve these issues without exposing this part of the runtime.

Issue #1085, opened 2018-07-14
Isomorphically rendered components cannot get the correct state.


ulivz:
### Details

Recently we upgrade `marko.js` from `4.7` to `4.12`, but our code doesn't work since then.

We all used isomorphically rendered UI component, so the component will be rendered at server, although the data returned to client is same to before, but due to this commit (https://github.com/marko-js/marko/pull/1010/commits/66b43e33c267029996539e48ca0e198c909fd1f6):

https://github.com/marko-js/marko/blob/b5f5f669ae2aff81868eafb8c46d466b61b6a6d3/src/components/ComponentDef.js#L130-L140

Since the input was an empty object, so it will cause the client cannot initialize the correct state.

### Expected Behavior

Client should get the correct initial state.

### Actual Behavior

Client cannot get the correct initial state.

### Possible Fix

I cannot commented these following code, all work for me.

``` js
 if (flags & FLAG_WILL_RERENDER_IN_BROWSER) { 
     if (component.onCreate) { 
         component.onCreate(input, { global: global }); 
     } 
     if (component.onInput) { 
         input = component.onInput(input, { global: global }) || input; 
     } 
 } else { 
   // Only remain this branch.
}
```

But I am sure it isn't the fix we want. and sorry I don't have too much time to continue.

### Your Environment

- Environment name and version (e.g. Chrome 39, node.js 5.4): node.js 6.11.3
- Operating System and version (desktop or mobile): all side.
- Link to your project: eBay's private project, sorry.

### Steps to Reproduce

1.  first...
2.
3.
4.

### Stack Trace

</details>

mlrawlings:
> Link to your project: eBay's private project, sorry.

@ulivz can you ping me on slack?  My userid is `mirawlings`

mlrawlings:
I think this is a duplicate of #1043 (see there for more context).  You can add the following tag to your component to get the old behavior:
```marko
<marko no-browser-rerender />
```

ianvonholt:
@mlrawlings This is problematic. When you start to serialize any sort of sensitive data within the input stream on the server, any references to input specific data with onCreate is then not available in the browser.

The only work around to this is creating a split component file on any components that you only want the onCreate to run server side only.

Example: 
```
class {
   onCreate (input) {
      this.state = {
        firstName: input.user.firstName,
        lastName: input.user.lastName
      }
  }

  onInput (input) {
    // This would cause the browser to throw an error in onCreate, but not the server.
    return {}
  }
}

[...]

Issue #1043, opened 2018-05-15
hydration not working alongside server-only onCreate


yomed:
### Details

@mlrawlings I'm using `onCreate` for server-only code to set the initial state of my component. As of `4.9.0`, the component does not get hydrated correctly, and `onCreate` gets fired in the browser. I can see this change was part of https://github.com/marko-js/marko/pull/1010, but I'm not sure if that missed use cases, or if I'm missing something.

If I comment out these lines (so that it goes to the `if(state)` part), everything works normally:
https://github.com/marko-js/marko/blob/7c5089b54c889cc8fc20c9f59eb0d591402e18b7/src/components/ComponentDef.js#L130-L137

### Expected Behavior

It seems that are some cases where `onCreate` should not be run in the browser, and instead the state should be taken from the serialization directly.

### Actual Behavior

`onCreate` runs in the browser instead of getting state through this code:
https://github.com/marko-js/marko/blob/7c5089b54c889cc8fc20c9f59eb0d591402e18b7/src/components/ComponentDef.js#L138-L153

### Possible Fix

<details><summary>Additional Info</summary>

### Steps to Reproduce

This bug is currently wrapped up in a complicated component, but I can try to create a reproducible example if needed.

</details>

mlrawlings:
@yomed The code you reference there is the difference in how a split component and non-split component are mounted in the browser.  

Is your component a stateful component that has server-only initialization code?

If a component has server-only code, you can make sure the component only runs `onCreate` on the server by splitting it and ensuring that it is not placed under another stateful component that will re-render and trigger the full lifecycle at some point, but because it's split and a root component, the template would not be send down for re-rendering.

---

This change to call `onCreate`/`onInput` at mount time could potentially be reverted and we might want to revisit it.  We document that `onMount` is the [first lifecycle method called in the browser](https://markojs.com/docs/components/#codeonmountcode), but this hasn't been true since [`marko@4.3.0`](https://github.com/marko-js/marko/pull/672).  

Stateful components re-render as part of mounting.  For top-level components, they were using the values from `onCreate`/`onInput` then re-rendering, but components nested under these top-level component went through their full lifecycle (including `onCreate`/`onInput`).

Data generated server-side in `onCreate`/`onInput` is regenerated in the browser and this means a few things: 
- if data can only be generated on the server, it will break in the browser.
[...]

PR #1289, opened 2019-03-15
add option to revert to old hydrate (pre 4.9) behavior

mlrawlings:
## Description

Marko 4.9.0 introduced a change to hydration that can cause issues (#1043, #1085).  But the recommended solution (`<marko no-browser-rerender />`) is intended to revert to Pre 4.3.0 behavior and causes a re-render not to happen.

This PR adds the following option to the marko tag.  It is not intended to be used long-term only as a stop-gap.   This PR also adds a deprecation warning for both this new option and `no-browser-rerender`.

```marko
<marko deprecated-no-create-or-input-for-top-level-hydrate /> 
```

## Checklist:

### DIFF
--- a/test/components-pages/fixtures/old-hydrate-no-create-no-input/tests.js
+++ b/test/components-pages/fixtures/old-hydrate-no-create-no-input/tests.js
@@ -1,0 +1,25 @@
+var path = require("path");
+var expect = require("chai").expect;
+
+describe(path.basename(__dirname), function() {
+    it("should mount, not call the onCreate/onInput methods and update", function() {
+        var component = window.testComponent;
+
+        expect(component.input.start).to.equal(10);
+        expect(component.state.count).to.equal(10);
+        expect(component.el.querySelector(".count").innerHTML).to.equal("10");
+
+        component.increment();
+        component.update();
+
+        expect(component.input.start).to.equal(10);
+        expect(component.state.count).to.equal(11);
+        expect(component.el.querySelector(".count").innerHTML).to.equal("11");
+    });
+
+    it("should still call all lifecycle methods on child components", function() {
+        expect(window.childCreateCalled).to.equal(true);
+        expect(window.childInputCalled).to.equal(true);
+        expect(window.childRenderCalled).to.equal(true);
+    });
+});
--- a/src/components/ComponentDef.js
+++ b/src/components/ComponentDef.js
@@ -7,10 +7,9 @@ var extend = require("raptor-util/extend");
 var KeySequence = require("./KeySequence");
 
 var FLAG_WILL_RERENDER_IN_BROWSER = 1;
-/*
-var FLAG_HAS_BODY_EL = 2;
-var FLAG_HAS_HEAD_EL = 4;
-*/
+// var FLAG_HAS_BODY_EL = 2;
+// var FLAG_HAS_HEAD_EL = 4;
+var FLAG_OLD_HYDRATE_NO_CREATE = 8;
 
 /**
  * A ComponentDef is used to hold the metadata collected at runtime for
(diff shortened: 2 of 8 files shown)


StarpTech:
Hi, could you describe more in depth why a user has to use it? Does it solve a bug? When yes why it's not just fixed? 
What's the difference between both modes?

test/components-pages/fixtures/old-hydrate-no-create-no-input/tests.js
```diff
@@ -0,0 +1,25 @@
+var path = require("path");
+var expect = require("chai").expect;
+
+describe(path.basename(__dirname), function() {
+    it("should mount, not call the onCreate/onInput methods and update", function() {
```
StarpTech: Where did you check that the methods wasn't called?

DylanPiercey: He added throws in the lifecycle (https://github.com/marko-js/marko/pull/1289/files#diff-809b6711d16268823b5b2c50c4366ed4R4) which would cause the test to fail.

mlrawlings:
@StarpTech We introduced a change in 4.9 that, while not a bug, probably should have been released as a major version.  While it didn't affect most apps, there have been a couple apps we've seen that rely on the old behavior.  Because this change was released as a minor version, we felt the responsible thing to do was to add an option to revert to the old behavior.

Prior to 4.9 top level components did not call onCreate and onInput when hydrating in the browser. onRender was the first lifecycle method and it used values serialized from the server.  However, any children of these top level components did go through the full lifecycle.  Now all components do.

This change is only intended to be a stop-gap.  If your app is already running after upgrading to 4.9+, this isn't something you should consider using.  If you do have any server-only code that needs to run in a component that renders client-side as well (has a `component.js` or `class` in the marko file), then it should be pulled up to a higher level.  

If you have something in `onCreate`, `onInput` or `onRender` that you only want to run on the server, it means you don't really need it in a lifecycle.  The server doesn't really have lifecycles, it just runs those three methods before rendering the template.  Anything that's server only and done in one of those lifecycles should just be done at the top of the corresponding marko file.

As an example:

```marko 
import fs from "fs";

class {
  onCreate(input) {
    this.state = { contents: fs.readFileSync(input.filename) };
  }
}

<div>
  ${state.contents}
</div>
```

This state isn't really "state".  It's never going to change.  And because you're using fs, you probably only intend this component to run on the server, but it will also be sent to the browser.  Prior to 4.9, if this were a top-level component, this would have been fine since the state value set in `onCreate` would be serialized and used instead of rerunning `onCreate`, but you're still sending an unnecessary template that will never actually rerender in the browser (and if your bundler automatically inserts a `fs` shim, that would be sent as well!)

Instead, we'd recommend this:

```marko
import fs from "fs";

$ const contents = fs.readFileSync(input.filename);

<div>
  ${contents}
</div>
```

Now, there's no associated `class` (or `component.js`), so this won't be sent to the browser (unless it's used by another component that is sent to the browser).  If the logic was more complex than this, we might also recommend moving it out into a separate js file and importing a helper function:

```marko
import fs from "fs";
import getData from "./get-data";

$ const data = getData(input);

<div>
  ${data.contents}
</div>
```

StarpTech:
@mlrawlings thank you for the detail explanation. You explanation brings me to the question: How can I prerender a stateful component on server-side without to toggle a rerender on client-side?

Imagine the scenarion with translations or with a component which only serves as a controller to delegate events to other components. Right now, due to the behaviour that the template is rerendered on client-side all my prerendered values are destroyed. This force me to provide a isomorphic solution for everything. Also `no-update` has no effect. 

https://github.com/marko-js/marko/issues/1291

ianvonholt:
The changes after 4.9 to the lifecycle methods really messed up the `onInput` method. We used it to serialize data on the server while still keeping component arguments to a minimum. 

Example: 

```
onInput (input) {
   return {
     firstName: input.user.firstName || null, 
     lastName: input.user.lastName || null
   }
}
```

In the above, the `input.user` object would be passed appropriately, but then input would only contain `firstName` and `lastName`. Thus, when run in the browser the component would have null values for 
`firstName` and `lastName`. Split components solved the problem for a lot of the more complex components we had, as it allowed us to define specifically what was run on the server, and what was run in the browser.

Issue #1291, opened 2019-03-15
Marko destroys hydrated page with a rerender on client side


StarpTech:
### Marko Version: 14.5.2

As titled. I have a stateful component (empty class) with a `<i18n/>` component which use [i18n](https://www.i18next.com/) as utility. When the page is hydrated on the client, it will be rerendered. This rerender destroys all my translated strings and results in empty strings because the utility isn't available on client side. Is marko not able to check if the DOM really needs to be rerendered? The template isn't connected to any state.

Another example:

Imagine the scenarion with translations or with a component which only serves as a controller to delegate events to other components. Right now, due to the behaviour that the template is always rerendered on client-side all my prerendered values are destroyed. This force me to provide a isomorphic solution for everything. Also no-update has no effect. How do you handle this?

There is compatibility tag `<marko no-browser-rerender />` which disable the rerender on client-side but I can't use it because then the `onCreate` isn't fired anymore on client-side and I can't handle component events in it.

StarpTech:
Marko supports hydration only on component level and based on the changes of two renderings. There isn't a concept of in-place updates of node values (text, attributes) which weren't change since the server DOM. This means that any stateful component has to provide client und browser implementations and since any child component of a stateful will be serialized to the client too, we also have to care about them as well.

ianvonholt:
I'm not sure if this is related, but I've been having a lot of state hydration problems with split components recently. Specifically, states defined within onCreate, on the server, will not be accessible on the browser for actual rendering. 

I'll see if I can't get a reproducible example up to illustrate.

StarpTech:
@ianvonholt this sounds like a different problem. Please create a seperate issue.

PR #1292, opened 2019-03-15
Fix regression for no-update on dynamic tag html attributes

DylanPiercey:
## Description

Currently if you have a dynamic tag that renders an html element with a `no-update` attribute the attribute will still be updated on new renders. This is a regression from https://github.com/marko-js/marko/blob/master/CHANGELOG.md#4110 (also happens for the include tag since it migrates to the dynamic tag).

This PR passes the properties (containing info like which attributes to preserve) through to the dynamic tag and resolves the issue.

## Checklist:

### DIFF
--- a/src/compiler/ast/CustomTag.js
+++ b/src/compiler/ast/CustomTag.js
@@ -619,12 +619,23 @@ class CustomTag extends HtmlElement {
                           [].concat(builder.parseExpression(this.argument))
                       )
                     : builder.literalNull();
+                const properties = this.getProperties();
                 tagVar = context.helper("dynamicTag");
                 tagArgs = [
+                    builder.identifierOut(),
                     this.tagNameExpression,
                     inputProps,
                     argumentNode,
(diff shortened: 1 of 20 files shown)


StarpTech:
Great catch! Does it fix https://github.com/marko-js/marko/issues/1291 as well ?

StarpTech:
Does it affect components inside dynamic attribute tags? I have the issue that my components are rerendered on client side even with zero state. Also the no-update attribute has zero impact.

DylanPiercey:
@StarpTech this does not solve #1291, been meaning to reply on that. `:no-update` on attributes in Marko does not currently work for components, only html elements. The weirdness is around the "dynamic tags" which can be html elements. In this case there was a regression where the `:no-update` attributes were not honored which this fixes.

StarpTech:
I tried to apply `no-update` on html elements inside components but it gets still rerendered. Can I support you in any way?

PR #1315, opened 2019-04-08
Corrected typos

lucas0707:
Corrected two types, from its to it's and cumbersum to cumbersome

## Checklist:

### DIFF
--- a/docs/state.md
+++ b/docs/state.md
@@ -1,6 +1,6 @@
 # State
 
-The output of a component is based on input properties passed from its parent as attributes. However, a component may also maintain internal state that it uses to control its view. If Marko detects a change to either input or to the internal state, the view will automatically be updated.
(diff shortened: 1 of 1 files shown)


docs/state.md
```diff
@@ -1,6 +1,6 @@
 # State
 
-The output of a component is based on input properties passed from its parent as attributes. However, a component may also maintain internal state that it uses to control its view. If Marko detects a change to either input or to the internal state, the view will automatically be updated.
+The output of a component is based on input properties passed from it's parent as attributes. However, a component may also maintain internal state that it uses to control its view. If Marko detects a change to either input or to the internal state, the view will automatically be updated.
```
austinkelleher: I think the previous grammar was actually correct here. I'm fairly confident that the correct usage here should be `"its"` instead of `"it's"`.

lucas0707: You are right! I reverted the change.

casebuiltfurniture:
Ok

austinkelleher:
Thanks for the fix @lucas0707!

PR #1332, opened 2019-05-15
Deprecated body-only-if tag

agliga:
## Description

Changed dynamic tags to take null and not render the outer tag, but still render the body of the tag
If the user does not want to render contents, they should use if tags

It removes an unneeded tag and makes the syntax more concise.

## Checklist:


docs/core-tags.md
```diff
@@ -206,14 +206,14 @@ _HTML Output:_
 
 ## `body-only-if`
```
DylanPiercey: I think we should move this whole section (sans the `body-only-if` heading) into a pro tip on the dynamic tag name syntax section: https://github.com/marko-js/marko/blob/master/docs/syntax.md#dynamic-tagname

src/core-tags/migrate/all-tags/body-only-if.js
```diff
@@ -0,0 +1,52 @@
...
+
+        node.removeAttribute("body-only-if");
+        const attributes = node.attributes;
+        const tagDef = node.tagDef || {};
+        const dynamicTag = builder.htmlElement(
+            undefined,
+            attributes,
+            node.body
+        );
+        let tagName =
```
DylanPiercey: might be cleaner to change to:

```js
const tagName =
  node.rawTagNameExpression ||
  (tagDef.html
    ? builder.literal(node.tagName)
    : importTag(tagDef.renderer || tagDef.template, context));
```

agliga:
retest

src/core-tags/migrate/all-tags/body-only-if.js
agliga: Done

docs/core-tags.md
agliga: Moved code

src/compiler/CompileContext.js
```diff
@@ -330,6 +330,17 @@ class CompileContext extends EventEmitter {
         return this._errors;
     }
 
+    getRelativePath(importPath) {
```
mlrawlings: `getRequirePath` should handle both relative paths and modules.  Do we need this new function?

src/core-tags/migrate/util/import-tag.js
```diff
@@ -19,13 +20,27 @@ module.exports = function importTag(importPath, context) {
...
+            }
+        }
+    });
+    walker.walk(context.root);
 
     let identifier = requestedTagName;
     let i = 1;
 
     while (
         migrateImports.identifiers[identifier] ||
```
mlrawlings: This check may be redundant now.  Previously we were looking at the original source, but now we we're checking all imports and scriptlets from the current ast.  When a new import identifier is added, the ast should reflect this the next time this check is run.

src/compiler/CompileContext.js
agliga: It does, however what happens is it tries to resolve the module and we have tests that are importing files that do not exist which fail.
I spoke to @DylanPiercey about this, he told me to add this function since the other tests were failing.

src/core-tags/migrate/util/import-tag.js
agliga: I removed it and the tests are timing out (looping indefinitely), I think this also helps if it has already been loaded it does not duplicate that line of code.

PR #1373, opened 2019-07-02
Fallback to parent component helpers when unable to find the legacy parent

DylanPiercey:
## Description

In Marko 3 it is possible to have a template which relies on the parent widget arbitrarily higher in the tree.
We account for this by also tracking the legacy component def tree, however is some cases the parent widget might have been converted into a Marko 4 component and that causes `w-x` attributes to break.

This PR updates the legacy widget parent lookup to fallback to a regular v4 component.

## Checklist:


DylanPiercey:
Opted to not go with this fallback since it could end up masking errors.

PR #1392, opened 2019-08-02
Update wiki URLs

tigt:
[I made some edits to wiki titles](https://gitter.im/marko-js/marko?at=5d43a91ea96def4751a0a5fd) before realizing their URLs were referenced here, so this fixes that.

### DIFF
--- a/docs/10-awesome-marko-features.md
+++ b/docs/10-awesome-marko-features.md
@@ -290,4 +290,4 @@ Coming soon: auto correction and autonomous coding
 ---
 
 _Cover image credit:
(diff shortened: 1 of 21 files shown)


DylanPiercey:
@tigt this is excellent, thanks for picking this up 😄.

If you could sign the CLA then I'm happy to merge!

tigt:
I did sign it after the message popped up — odd that it hasn't updated yet.

When I return to the CLA page it says I agreed and the form is disabled, so
I can't really try it again

DylanPiercey:
Thanks again for the PR, not sure what's up with the CLA 🤷‍♂.

PR #1415, opened 2019-09-04
Fix widget args not preserved when nested tags used

DylanPiercey:
## Description

This is a patch for Marko v3 which fixes an issue where nested tags would cause the event handlers (and any thing else related to the internal widget args) to be skipped if the nested tag itself also had widget args.

## Checklist:


mlrawlings:
Marko 3 is dead.

PR #1424, opened 2019-10-03
Deprecated @key and * without target-property

agliga:
## Description
* Added `@key` as deprecated (if there's a better way to check if the key is not part of a native tag, let me know. I had some tests fail because of that)
* Made missing `target-property` have deprecation warning
* Changed all the `marko.json` files to not use the old syntax

* All the tests that are here should pass, except for the ones that were failing from the new-compiler-integration branch.

## Checklist:


src/taglib/taglib-loader/Tag.js
```diff
@@ -107,6 +112,12 @@ class Tag {
                 ) {
                     attr.targetProperty = null;
                 } else if (!attr.targetProperty) {
+                    complain(
+                        "@* target-property property required. The recommended value is to set target-property to null.",
```
DylanPiercey: I would make the complain something like:

```
The default "targetProperty" for "@*" attribute definitions is changing from "*" to "null" (merged in with the rest of the input) in a future Marko release. In order to avoid an issue upgrading, please explicitly define the "targetProperty".
```

src/taglib/taglib-loader/Tag.js
```diff
@@ -269,6 +280,10 @@ class Tag {
         return (this._nodeFactory = nodeFactory);
     }
 
+    isCustomTag() {
+        return this.renderer || this.template || this.isNestedTag;
```
DylanPiercey: Not sure how I feel about this function, `nestedTags` are not custom tags :p.

Not sure if we'll still need this if you switch to checking for the specific taglib id defining the `key` attribute though.

test/taglib-lookup/fixtures/declared-and-dynamic-attrs/marko.json
```diff
@@ -2,6 +2,9 @@
     "<test-dynamic-attribute>": {
         "renderer": "./renderer.js",
         "@foo": "string",
-        "@*": "boolean"
+        "@*": {
```
DylanPiercey: I'm surprised that none of these changes broke any tests, are none of them relying on `input["*"]`?

agliga: Yep, they did not fail. 
This seems like it's not a real good test (and the template had a TBD in it from 3 years ago):
```
    var attr = lookup.getAttribute("test-dynamic-attribute", "DYNAMIC");
    expect(attr != null).to.equal(true);
    expect(attr.type).to.equal("boolean");
    expect(attr.name).to.equal("*");

    attr = lookup.getAttribute("test-dynamic-attribute", "foo");
    expect(attr != null).to.equal(true);
    expect(attr.type).to.equal("string");
```

PR #1398, opened 2019-08-11
WIP: Remove warning for XML namespaces on attributes

tigt:
Fixes #1367, #1297

I’m having a hard time understanding what exactly is causing the 2 final unexpected errors:

```
1) compiler/fixtures-html
       xml-namespaced-tag
     Error: Generating code for TemplateRoot node failed. Error: Error: Generating code for <OpenSearchDescription> tag failed. Error: Error: Invalid JavaScript identifier: xmlns:moz
      at module.exports (node_modules/raptor-util/createError.js:7:50)
      at CodeGenerator._invokeCodeGenerator (src/compiler/CodeGenerator.js:136:19)
      at CodeGenerator._generateCode (src/compiler/CodeGenerator.js:231:38)
      at CodeGenerator.generateCode (src/compiler/CodeGenerator.js:282:14)
      at Compiler.compile (src/compiler/Compiler.js:186:38)
      at _compile (src/compiler/index.js:91:35)
      at Object.compileFile (src/compiler/index.js:158:16)
      at runTestForExtension (test/compiler/html.test.js:57:36)
      at Context.done (test/compiler/html.test.js:76:28)
      at Context.<anonymous> (test/autotest.js:145:31)
  Caused by: Error: Generating code for <OpenSearchDescription> tag failed. Error: Error: Invalid JavaScript identifier: xmlns:moz
      at module.exports (node_modules/raptor-util/createError.js:7:50)
      at CodeGenerator._invokeCodeGenerator (src/compiler/CodeGenerator.js:136:19)
      at CodeGenerator._generateCode (src/compiler/CodeGenerator.js:231:38)
      at node.forEach.child (src/compiler/CodeGenerator.js:149:26)
      at ArrayContainer.forEach (src/compiler/ast/ArrayContainer.js:18:26)
      at CodeGenerator._generateCode (src/compiler/CodeGenerator.js:147:18)
      at CodeGenerator.generateCode (src/compiler/CodeGenerator.js:282:14)
      at TemplateRoot.generateCode (src/compiler/ast/TemplateRoot.js:42:29)
      at CodeGenerator._invokeCodeGenerator (src/compiler/CodeGenerator.js:116:29)
      at CodeGenerator._generateCode (src/compiler/CodeGenerator.js:231:38)
      at CodeGenerator.generateCode (src/compiler/CodeGenerator.js:282:14)
      at Compiler.compile (src/compiler/Compiler.js:186:38)
      at _compile (src/compiler/index.js:91:35)
[...]

PR #1431, opened 2019-10-19
Copyediting

tigt:
I wasn’t sure how to handle the array-like code sample since `Object.entries()` seems to fit the use-case much better, so I just deleted it.

• More semantic/accessible code examples
• Don’t nest `<h1>` inside `<p>`
• Mostly more succinct wording, spelling, and grammar-checking

## Checklist:

- [x] ~~I have added tests to cover my changes.~~


docs/core-tags.md
```diff
@@ -1,434 +1,388 @@
...
+Instead, preserve whitespace with the `preserve-whitespace` attribute:
 
 ```marko
-<div marko-preserve-whitespace>
-    All of this
-    whitespace   will
-    be preserved.
-</div>
-```
+<pre marko-preserve-whitespace>
```
mlrawlings: Since `<pre>` will preserve whitespace by default, `marko-preserve-whitespace` is redundant here.  But `<div>` alone also isn't a great example because preserving the whitespace isn't going to matter.  Perhaps we should include a style block that gives the tag `white-space:pre`:

```marko
style {
    .poem {
        white-space: pre;
    }
}

<div.poem>
    who ate all the cookies??
                   I     did.
</div>
```

tigt: Excellent point. I’ve been trying to think of an example where `marko-preserve-whitespace` would fit well on a known, semantic use-case, but it looks like all the elements that want that behavior have it built-in (`<pre>`, `<textarea>`, `<script>`, etc.)

I don’t want to use a `<div>` if only because concrete poems, semantically, _should_ be `<pre>` — if Marko didn’t discourage generation of XML (#1398), something involving `xml:space="preserve"` might work…

Other examples I can think of similarly shouldn’t be encouraged — `<xmp>`, `<plaintext>`, `<listing>`…

I know this seems trivial, but I think docs are one of the best ways to teach/encourage good HTML, so I care about it maybe more than I should. What use-cases was `marko-preserve-whitespace` added for in the first place?

tigt: I propose we punt on this because without a real use-case for why `marko-preserve-whitespace` exists, I’m not sure the docs backing it up is that important?

mlrawlings: Yeah, I'd like to get the rest of the changes in... 

What about this as an example? 
```html
<style>
  code {
    white-space: pre;
  }
</style>

<code marko-preserve-whitespace>
  import { foo } from "bar";
  console.log(foo);
</code>
```

PR #1505, opened 2020-02-20
Deprecate JSON.stringify for attributes and other attr string fixes

DylanPiercey:
## Description

This PR aims to address some incompatibilities between Marko 3 and 4 with the compatibility API and deprecates using `JSON.stringify` to stringify attribute values which are objects.

The main case where `JSON.stringify` was meant to be used was by the runtime with `data-marko` output by split components on the server side. This is now handled by a standalone helper.

This PR includes a small breaking change which should not effect any users which is that it will no longer call `JSON.stringify` on non native objects and instead calls `toString`. This is primarily done to improve compatibility with Marko 3.

Previously the non standard template string syntax `<div x="${something}">` would output string concat, however this was not entirely safe. Now this is migrated to pass through the value as is in this case.

Finally `toHTML` is no longer called for server side attribute values. This doesn't make sense and the behavior already is not supported in the vdom renderer.

## Checklist:


tigt:
+1 to this; it makes using arrays for comma-separated attributes (`srcset`, `sizes`, `accept`, etc.) slightly easier, and matches the behavior of passing an array to `.setAttribute`.

Issue #1457, opened 2019-12-02
Performance - use JSON.parse while sending the json to browser.


pajaydev:
### Description

From recent talks from v8 team, we know that `JSON.parse approach is much faster compared to the JavaScript object literal. JSON.parse() is much faster than JavaScript evaluation`. I got inspiration from the below article to make this change
https://v8.dev/blog/cost-of-javascript-2019#json

### Why

This may be effective when you use handful of stateful components in your marko app and marko ships those to the browser. In the above case, browser takes time to parse and compile JS, I have made a small comparison with a sample JSON file in my machine. Below are the results

#### Before changes
<img width="1417" alt="Screen Shot 2019-12-01 at 9 31 29 PM" src="https://user-images.githubusercontent.com/21375014/69934518-fa8ded80-1486-11ea-9494-dfc8b147ee93.png">

#### After changes
<img width="1417" alt="Screen Shot 2019-12-01 at 9 31 16 PM" src="https://user-images.githubusercontent.com/21375014/69934525-fe217480-1486-11ea-8ab0-e18927aea9d9.png">

### Possible Implementation & Open Questions
We know that marko uses `warp10` internally to serialize the data while sending it to browser. We can make this change as suggested below in this file https://github.com/marko-js/marko/blob/master/src/runtime/components/index.js#L188

```javascript
      componentGlobalKey +
        "=(window." +
        componentGlobalKey +
        "||[]).concat(JSON.parse(" +
        JSON.stringify(safeJSON(warp10.stringify(renderedComponents))) +
        "))||" +
        componentGlobalKey
```

Kindly share any thoughts, concerns or feedback.

CC: @DylanPiercey @mlrawlings 

### Is this something you're interested in working on?

Yes

tigt:
Related: https://github.com/marko-js/warp10/issues/2

tigt:
Slightly related: it should be safe to replace the `Array#concat()` call in your suggested code with `Array#push()`, right? 2 less characters, but more importantly it won’t return a newly-copied array.

DylanPiercey:
@tigt after Marko initializes it actually replaces the global key with an object which has a `concat` method. Since the scripts generated also assign to the global, we also return the `concat` object from the `concat`.

See here: https://github.com/marko-js/marko/blob/master/src/runtime/components/init-components-browser.js#L275

DylanPiercey:
[...]

PR #1551, opened 2020-04-17
perf: serialize data as strings, reduce bytes per flush

DylanPiercey:
## Description

This PR makes two main optimizations:

1. Instead of serializing component initialization data as JavaScript, it instead puts it inside a JavaScript string.
2. Reduces the number of bytes per flush by using string concat and `JSON` instead of `window` as our global of choice 🤭.

fixes https://github.com/marko-js/marko/issues/1457

## Details

Why `JSON` instead of `window`? It is fewer bytes and has less chance of conflicting with other global variables. We sorted the available global keys with browser support of at least IE10 and `JSON` is one of the shortest available namespaces and it somewhat aligns with what is happening. 🤷‍♂️ 

The global added is now a setter as well which still allows for the immediate initialization of components. Previously we returned a fake array with a `concat` method.

## Checklist:


packages/marko/src/runtime/components/init-components-browser.js
```diff
@@ -252,41 +251,38 @@ function initClientRendered(componentDefs, doc) {
...
+      set: function(v) {
+        initServerRendered(JSON.parse(v.slice(10)), doc, runtimeId);
+      }
     });
 
-    if (renderedComponents && renderedComponents.forEach) {
-      renderedComponents.forEach(function(renderedComponent) {
-        fakeArray.concat(renderedComponent);
+    if (renderedComponents) {
+      JSON.parse("[" + renderedComponents.slice(10) + "]").forEach(function(
```
tigt, line 275: I like this — it looks like in the future it could even throttle itself based on how much distress the user’s CPU is in

packages/marko/src/runtime/components/index.js
```diff
@@ -10,11 +10,16 @@ var FLAG_HAS_RENDER_BODY = 2;
...
-  } else {
-    return "\\u" + match.charCodeAt(0).toString(16);
+function safeStringJSONReplacer(match) {
+  switch (match) {
+    case "'":
+      return "\\'";
+    case "\\":
+      return "\\\\";
+    case "</script":
+      return "\\u003C/script";
```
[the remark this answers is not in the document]
DylanPiercey: I did some cross browser testing for spaces in the closing script and all are pretty good about it. For some reason I didn’t test spaces before the closing angle bracket. I’ll test that and update this if there is no issue there.

packages/marko/src/runtime/components/init-components-browser.js
DylanPiercey: Yeah that was one of the things I liked as well, we can lazily parse the JSON.

packages/marko/src/runtime/components/index.js
```diff
@@ -10,11 +10,16 @@ var FLAG_HAS_RENDER_BODY = 2;
...
+    default:
+      return "\\u" + match.charCodeAt(0).toString(16);
```
tigt, line 22: Does this output leading zeroes if the `charCode` is less than 4 digits? Or do you not need to bother, since the characters it catches will only end up at those higher codepoints anyway?

packages/marko/src/runtime/components/index.js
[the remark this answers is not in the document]
tigt: I double-checked the JSON spec and [it _looks_ like `</script` can be escaped as the slightly shorter `<\\/script`](https://i.stack.imgur.com/SHLOB.gif), instead of `\\u003C/script` — it also looks like trailing garbage before the final `>` doesn't change how it closes the parent opening script tag, so that's not a concern either.

packages/marko/src/runtime/components/init-components-browser.js
```diff
@@ -252,41 +251,38 @@ function initClientRendered(componentDefs, doc) {
...
 
   if (type !== "object") {
     if (type === "string") {
       runtimeId = renderedComponents;
-      globalKey += runtimeId + "_C";
+      globalKey = runtimeId;
     } else {
-      globalKey += (runtimeId = DEFAULT_RUNTIME_ID) + "C";
+      globalKey = runtimeId = DEFAULT_RUNTIME_ID;
     }
```
tigt, line 264: What do you think about shortening this to:

```js
if (type === "string") {
  runtimeId = renderedComponents;
} else {
  runtimeId = DEFAULT_RUNTIME_ID;
}

globalKey = runtimeId;
```

Or maybe even:

```js
globalKey = runtimeId = (
  type === "string" ? renderedComponents : DEFAULT_RUNTIME_ID
)
```

DylanPiercey: Yeah good catch. I realized that one as well. With the changes I've made `globalKey` and `runtimeId` are one and the same. I think I'll just remove `globalKey`.

packages/marko/src/runtime/components/index.js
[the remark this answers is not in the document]
DylanPiercey: Ah this is great! Thanks for doing some research on this one. To be clear are you saying `</script abc>` does not get parsed in any browser as the closing script?

packages/marko/src/runtime/components/index.js
DylanPiercey: We only escape two characters both with 4 digits [\u2028|\u2029](https://github.com/marko-js/marko/pull/1551/files#diff-55c5215fd78cb23b9a4cb49e164d0066R4). Having said that I think this would be clearer to just include the full escaped sequence, ie:

```
      case "\u2028":
        return "\\u2028";
      case "\u2029":
        return "\\u2029";
```

packages/marko/src/runtime/components/index.js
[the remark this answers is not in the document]
tigt: Unfortunately, the opposite — it gets parsed _everywhere_ according to spec as the closing script.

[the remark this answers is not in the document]
DylanPiercey: Lame, okay then. We'll leave it as is, but incorporate your `<\\/script` suggestion.

packages/marko/src/runtime/components/index.js
tigt: That makes more sense to me upon reading it. Might even be faster too — possibility of a `Map` instead of a `switch` statement too, I think.

DylanPiercey: I haven't benched `Map` vs `switch`, but I could only assume `switch` is faster? Am I wrong?

DylanPiercey: Hmm maybe I'm wrong: https://jsperf.com/if-switch-lookup-table/113 (tested in Chrome). 

Will need to do a more specific bench than this, but if it is similar perf it would make the code a bit nicer I think.

tigt: Honestly, at this point, I’ve been burnt too many times before on my guesses for what’s faster inside JavaScript engines.

DylanPiercey: Yeah the micro benchmarks lie quite often. Always have to do a real bench. At least for SSR we can get away with just benching a few node versions.

packages/marko/src/runtime/components/init-components-browser.js
DylanPiercey: Done! Realized I could shorten it further, since really the `type` here would either be `string` or `undefined`. I just made it `runtimeId = renderedComponents || DEFAULT_RUNTIME_ID`.

The API for `init` should be one of `init()`, `init("runtimeID")` or `init(serializedComponentDataObject)`.

packages/marko/src/runtime/components/index.js
DylanPiercey: Done!

tigt:
Circling back to this after some discussions in the Marko Discord about [using SubResource Integrity for Content Security Policies](https://github.com/marko-js/community/pull/2).

This PR currently serializes script data as:

```html
<script>JSON.M+=',{"l":1,"w":[["s0-0-5-5",0,{},{"f":1}]],"t":["/marko-test$1.0.0/render/fixtures/components-await-title/components/hello/index.marko"]}'</script>
```

For statically-generated hashes, it could look more like:

```html
<script
  data-s=',{"l":1,"w":[["s0-0-5-5",0,{},{"f":1}]],"t":["/marko-test$1.0.0/render/fixtures/components-await-title/components/hello/index.marko"]}'
  integrity="sha256-CSXorXvZcTkaix6Yvo6HppcZGetbYMGWSFlBw8HfCJo="
>JSON.M+=document.currentScript.data.s</script>
```

DylanPiercey:
> Circling back to this after some discussions in the Marko Discord about [using SubResource Integrity for Content Security Policies](https://github.com/marko-js/community/pull/2).
> 
> This PR currently serializes script data as:
> 
> ```
> <script>JSON.M+=',{"l":1,"w":[["s0-0-5-5",0,{},{"f":1}]],"t":["/marko-test$1.0.0/render/fixtures/components-await-title/components/hello/index.marko"]}'</script>
> ```
> 
> For statically-generated hashes, it could look more like:
> 
> ```
> <script
>   data-s=',{"l":1,"w":[["s0-0-5-5",0,{},{"f":1}]],"t":["/marko-test$1.0.0/render/fixtures/components-await-title/components/hello/index.marko"]}'
>   integrity="sha256-CSXorXvZcTkaix6Yvo6HppcZGetbYMGWSFlBw8HfCJo="
> >JSON.M+=document.currentScript.dataset.s</script>
> ```

When I take a look at this for Marko 6 I need to remember your solution here. In theory, by using an attribute to encode the JSON we should have better server side perf since escaping HTML quotes is less expensive than escaping script content.

DylanPiercey:
In our benchmarks there wasn't enough of a performance difference to justify this change at the moment.
I do still think there is a path forward here, perhaps following what @tigt had above, but I think it will be addressed in a different PR. https://github.com/marko-js/marko/issues/1457 will remain open.

Issue #1432, opened 2019-10-19
Server-rendered HTML minification


tigt:
## Description

Marko’s compilation to server-side templates could perform provably-safe HTML minification during compilation. (Runtime minification is probably too expensive to be worth it.)

## Why

Perf matters! Running Marko’s current home page through [kangax’s minifier](https://kangax.github.io/html-minifier/) saves 2,551 bytes. Not too bad after compression, but it’s still overhead, and fewer characters speed up parsing.

<details>
I used the relative URLs option, kept comments because I suspected Marko uses them for rehydration, and turned off the Minify CSS/JS options.
</details>

Existing HTML minifiers _can_ be plugged into a Marko app’s render pipeline, but they require buffering the stream, defeating any performance improvements.

## Possible Implementation & Open Questions

In development mode, probably don’t bother running this compilation step. But in production…

+ Normalize output for better compression
  - Normalize known tag/attribute name casing (might do this already?)
  - Sort attributes (probably alphabetically, but frequency would be even better)
  - Sort properties in `style`
  - Sort tokens in `class`, `rel`, and other TokenLists/similar
+ Collapse whitespace (Marko largely already does this, but it’s worth reexamining to ensure its behavior is as unsurprising and efficient as it could be)
+ Collapse all whitespace characters to a single space except when `<pre>`, `marko-preserve-whitespace`
+ Omit the closing slash on known void elements
+ Optional tags
  - Omit optional opening tags without attributes like `<head>` or `<body>`
  - Omit optional closing tags where safe — `</body></html>`, consecutive `<p>`s, `</option>` since `</select>` autocloses it
+ Omit attributes with default values :`method="get"`, `<input type="text">`, `<script type="text/javascript">`, etc. — this would also make rendering client-side slightly more efficient
+ Attribute values
  - Truthy booleans should omit `=""`
  - Literals not containing whitespace should omit quotes
  - Literals containing `"` or `'`' should use the other delimiter for the value
  - If the attribute value is known not to stringify with forbidden characters (`Number`, for example), it should omit quotes
  - Everything else should fall back to wrapping double quotes
+ Reencode character references
  - Non-HTML syntax characters like `&rarr;` should be dereferenced
  - Unnecessary encoding should be dereferenced
    * Unambiguous ampersands
    * Quotes and `>` outside opening tag contexts
    * `=` outside attribute values
  - Some character references could be reencoded to be shorter:
[...]

PR #1557, opened 2020-04-21
perf: minify runtime comments, remove unnecessary attr quotes

DylanPiercey:
## Description

Working toward: https://github.com/marko-js/marko/issues/1432

## Checklist:

This PR removes unnecessary characters from the html output where possible:

1. Removes `--` from comments since all tested browsers do not require them, for example `<!M>` is a treated valid comment.
2. Removes unnecessary quotes around attributes where possible. Any attribute that doesn't contain `'">` or any space characters can have the quotes removed. Technically the [spec](http://w3c.github.io/html-reference/syntax.html#syntax-attr-unquoted) says that `=`, `<` and backticks also need to be escaped, however browsers appear happy to ignore that. The implementation could do better by ensuring the minimum number of escapes by switching between single or double quotes, but for performance reasons instead it will immediately use single quotes when the first double quote is found which works well for most cases.


redonkulus:
@DylanPiercey FYI, this ended up breaking marko snapshots locally since all the attrs were removed. I thought it was a bug in my code somewhere. It probably didn't warrant a major version bump, but can definitely cause builds to fail until updated.

DylanPiercey:
@redonkulus at the end of the day a snapshot test is going to capture many implementation details of Marko that we don't consider a "public api". We'd expect snapshots to need updating across many releases in Marko 4. The point of snapshot tests though are not to assert that something is working, but to let you know that something has changed. I think developers having to update snapshots during changes like this is acceptable.

tigt:
Removing the `--` delimiters from comments produces a lot of noisy errors/warnings from tools like the HTML validator for our automated & manual checks. Not sure how much that matters to you.

mlrawlings:
@tigt We may need to revisit that... 

It technically violates the spec but we tested it in Safari, FF, Chrome, Edge, and IE and it parses as expected.

tigt:
Yeah, I’m not aware of any HTML parsers that would refuse to work with it.

DylanPiercey:
It's tricky. Basically the shorter comments are objectively better since it's just less rendered and shipped code, but tools are going to complain.

I think Marko needs to take a stance on if it is more valuable to have output which linters like, or that is more performant / smaller.

mlrawlings:
We're reverting the comment change for now.  We'll revisit this in the future.

alexnaidovich:
Is there the way to bring the quotes for attributes' values (that were considered there to be unnecessary) back to the output? There **are** real cases when quotes are necessary. Maybe, are there some configurable options for that?

PR #1562, opened 2020-04-27
feat: add new import dynamic tag syntax to migrations

biaspro:
## Description

Update fixtures in migrations to use the new import dynamic component syntax

## Checklist:


packages/marko/src/core-tags/migrate/util/import-tag.js
```diff
@@ -7,19 +7,31 @@ module.exports = function importTag(baseImportPath, context) {
...
+    let fileName;
+    if (importPath[0] === "<") fileName = importPath.slice(1, -1);
+    else {
+      const match = nameReg.exec(importPath);
+      fileName = (match && match[1]) || "template";
+    }
+    return createTagName(fileName);
+  };
+
   const importPath = context.getRelativePath(baseImportPath);
```
DylanPiercey, line 28: I just realized: `importPath` on this line (used as a cache key below) would now be passing the shorthand such as `<component-a>` to the `getRelativePath` helper. I don't think we want that in this case, instead it should just use the `baseImportPath` as is.

packages/marko/src/core-tags/migrate/util/import-tag.js
```diff
@@ -1,25 +1,38 @@
...
     paths: {}
   });
-  const importPath = context.getRelativePath(baseImportPath);
-
-  if (migrateImports.paths[importPath]) {
-    return migrateImports.paths[importPath];
-  }
-
-  const match = nameReg.exec(importPath);
+  let isImportedAsTag = importIdentifier[0] === "<";
```
DylanPiercey, line 10: Minor nitpick, but we should probably make this a `const` to be consistent.

biaspro: ah yes good catch

PR #1563, opened 2020-04-28
Fix excludePackage because of window path confusion

cc-ebay:
On window platform `marko` has problems with paths. This PR fix the `excludePackage` function

## Description

Adding [upath](https://www.npmjs.com/package/upath) to normalize the path names.

## Checklist:

### DIFF
--- a/compiler/taglib-finder/index.js
+++ b/compiler/taglib-finder/index.js
@@ -16,7 +16,7 @@
 'use strict';
 
 var taglibLoader = require('../taglib-loader');
-var nodePath = require('path');
+var nodePath = require('upath');
 var fs = require('fs');
 var lassoPackageRoot = require('lasso-package-root');
 var resolveFrom = require('resolve-from');
--- a/package.json
+++ b/package.json
@@ -51,7 +51,8 @@
     "raptor-util": "^2.0.1",
     "resolve-from": "^1.0.0",
     "strip-json-comments": "^2.0.1",
-    "try-require": "^1.2.1"
+    "try-require": "^1.2.1",
+    "upath": "^1.2.0"
   },
   "devDependencies": {
     "bluebird": "^2.9.30",


cc-ebay:
I do not think I broke the test here: 
```
 1) render
       autotest
         [marko-compiler-options-preserve-comments] :
      AssertionError: '<div><!-- This comment should be preserved --> Hello World!</div><div><!--This comment should be preserved--> Hello World!</div deepEqual '<div><!-- This comment should be preserved --> Hello World!</div><div><!--This comment should be preserved-->Hello World!</div>
      + expected - actual
      -<div><!-- This comment should be preserved --> Hello World!</div><div><!--This comment should be preserved--> Hello World!</div>
      +<div><!-- This comment should be preserved --> Hello World!</div><div><!--This comment should be preserved-->Hello World!</div>
      
      at compareHelper (test/autotest.js:47:12)
      at Object.compare (test/autotest.js:55:13)
      at run (test/render-test.js:55:29)
      at autoTest (test/autotest.js:59:5)
      at Context.<anonymous> (test/autotest.js:83:21)
```

DylanPiercey:
Looking at this again I don't see any benefit that would be brought in by using `upath`.

The main issue that can happen with nodes built-in `path` is when the formats get mixed and combined, eg:

```js
path.join(someWindowsPath, "./this-is-a-unix-style-path")
```

We are careful in the codebase to avoid this. Although I have found one specific place in the codebase where this may be causing an issue, it would not have been fixed in this PR (I am putting up a separate PR).

We are happy to fix specific cases where there is a mistake made in using the `path` API with unix style paths, when everything should be generic and `path.sep` should be used if necessary. The alternative would be using `upath` throughout the entire codebase which I don't think is worth it over just fixing incorrect usages of the `path` api.

PR #1593, opened 2020-08-07
fix: set default value for textarea when created

DylanPiercey:
## Description

Currently, when actualizing a viritual `textarea` element (IE not server-rendered) we do not properly assign the `defaultValue` property. 

This PR sets both `defaultValue` and `value` in this case.

## Checklist:

### DIFF
--- a/packages/marko/src/runtime/vdom/VElement.js
+++ b/packages/marko/src/runtime/vdom/VElement.js
@@ -190,7 +190,7 @@ VElement.prototype = {
       }
 
       if (tagName === "textarea") {
-        el.value = this.___value;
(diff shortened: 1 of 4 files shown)


tigt:
Might this fix https://github.com/marko-js/website/issues/40?

DylanPiercey:
@tigt I do believe it would fix that.

triptych:
We need this for our `buyermanagementview` codebase as once we updated ebay-ui as a test we found a regression around the defaultValue attribute.

DylanPiercey:
@triptych this patch is released in `4.22.6`.

PR #1598, opened 2020-09-03
Hoist Static VDOM Node Creation

ryansolid:
This PR hoists static VDOM node creation so that they are not repeatedly created at runtime.

## Description

This change to the compiler looks for static native elements with static attributes and children and hoists portions of VDOM node creation out of the render function. This way they can be created once and reused. This improves rendering performance. 

This behavior was present in Marko 4 and was not yet implemented in Marko 5.

## Checklist:


packages/translator-default/src/analyze.js
```diff
@@ -0,0 +1,54 @@
...
+
+export const staticNodes = new WeakSet();
+
+export const visitor = {
+  MarkoText(path) {
+    staticNodes.add(path.node);
+  },
+  MarkoPlaceholder(path) {
+    const { confident } = path.get("value").evaluate();
+    if (confident && path.node.escape) staticNodes.add(path.node);
```
DylanPiercey, line 12: Probably best to guard the work of `path.evaluate` with a `!path.node.escape` check since that method is semi expensive.

packages/translator-default/src/analyze.js
```diff
@@ -0,0 +1,54 @@
...
+  },
+  MarkoTag: {
+    enter(path) {
+      // check if a for
+      // needed to handle global keys on elements that don't have specific key attributes
+      if (path.node.name.value === "for") path.skip();
```
DylanPiercey, line 18: Shouldn't `while` be considered here as well? And maybe some other control flow tags?

DylanPiercey (commented):
I'm a bit concerned that we are `path.evaluate`ing nodes multiple times (eg attributes). I'm not sure if it is cached on the Babel end by I'm pretty sure it's an expensive op.

Also I'm a bit concerned about all of this additional nested traversing we are doing. If there isn't a better way, we should probably consider turning that off in dev mode for build perf reasons.

ryansolid:
> Also I'm a bit concerned about all of this additional nested traversing we are doing. If there isn't a better way, we should probably consider turning that off in dev mode for build perf reasons.

Yeah this came about since you need to know about because a node must check it's children to know if it is static, but the code generation is different for children under the top-most static node. So you don't know if the parent is static until you look at the children and you don't know that the children should be processed a certain way until the parent has looked at all its children. It's possible if we used the chaining syntax everywhere this could be avoided but without changing how write works everywhere this is where I landed.

packages/translator-default/src/analyze.js
ryansolid: This is due to implicit keys under loops that break when hoisted. It looks like we use these keys not just top level in the loop. "while" makes sense are there others this applies to? For was the only one in the tests so it was the only one I hit issues with.

Issue #1624, opened 2020-11-08
Add a load option forceRecompile to force loader recompile template instead of using cache


lanyusan:
### Description

Support property `forceRecompile` of `options` argument  of marko loader `function load(templatePath, templateSrc, options) ` and update loader code to check if the property, `forceRecompile` is set to `true` , then recompile template,  i.e. `if( options && options.forceRecompile) recompile else useCache` 

### Why

I have a project that in a safe environment, internal web authors would use a web editor to edit marko template, primarily look and feel, static html and some basic template functions, save the content, then a preview server of marko engine would load the updated template from web, recompile it and generate preview.

Now the current marko loader always checks cache by file path and name and load compiled template if the template file was already compiled before even if the template file has been modified. So updated template **wouldn't** be rendered when refreshing preview page.

My use case is a production server for marko template authors to edit templates through web editor and update preview by refreshing preview page. So it is **NOT** an option to use `browser-refresh`. 

My preview server knows if a marko tempate at remote location has been modified. It would retrieve it and then force marko loader to recompile the marko template when user refresh preview page if this feature is implemented.

### Possible Implementation & Open Questions

Update loader functions to check `options` argument's `forceRecompile` property and skip loading from cache if set true.

### Is this something you're interested in working on?

**I have developed and tested it and would be happy to submit a PR.**

DylanPiercey:
I think my main issue with this is that in a lot of cases when you load a compiled template, many other templates get loaded as well. I think a simpler approach here would just be to manually invalidate the require cache. You can run something like this:

```js
for (const file in require.cache) {
  if (file.endsWith(".marko")) {
      delete require.cache[file];
  }
}
```

PR #1625, opened 2020-11-09
Added support for `forceRecompile` for marko loader. Marko issue #1624

lanyusan:
## Description

Support `options` parameter property `forceRecompile` for marko loader as defined by `module.exports = function load(templatePath, templateSrc, options) `

If set to true, marko loader will ignore cache and recompile template even it has been compiled previously
 

1. Check `forceRecompile` property of options argument.
2. If set to true, skip node js cache: 
```
  // Short-circuit loading if the template has already been cached in the Node.js require cache
  // However, if forceRecompile, then no short-circuit
  var cached = (options && options.forceRecompile) ? undefined : require.cache[templatePath];
```
3. If set to true skip getPreviousTemplate

```
    // Skip getting previous template if forceRecompile
    template = (options && options.forceRecompile) ? undefined : getPreviousTemplate(templatePath, options);
```

This will allow a template render engine to programmatically load template from remote template server and recompile the template without the need to restart node to invalidate cache. 

More detailed use case please refer to issue #1624

## Checklist:

This testing requires a separate setup of development environment. It has been tested with Fastify and point-of-view marko plugin, which isn't included in the PR as it is too heavy.
It has passed existing test cases.


DylanPiercey:
See my comment here https://github.com/marko-js/marko/issues/1624#issuecomment-781439225

PR #1638, opened 2020-12-18
feat: include input as a binding when crawling

DylanPiercey:
## Description
Currently babel sees `input` as a global variable when analyzing the scopes of the ast.
This PR makes it so that we register `input` as a normal `const` binding, allowing us to track references more easily.

Also fixes an issue where tag variable identifiers were not marked as referenced.


DylanPiercey:
My hopes were dashed and some of babel's utils don't play well with this idea.

PR #1657, opened 2021-02-05
Fix: Allow attribute tags inside of else-if tags

Khauri:
## Description

Fixes a bug where you aren't able to use attribute tags inside of an `<else-if>` tag.

Adds else-if to TRANSPARENT_TAGS Set in @marko/babel-utils

## Checklist:


DylanPiercey:
Awesome, thanks for the help with this @Khauri!

Issue #874, opened 2017-10-02
New Structure for Docs


mlrawlings:
### Guides

- Rendering
   - server side rendering
- Syntax
   - tags
   - attributes
   - inline javascript
- Loops & conditionals
   - keys
- Custom Tags
   - macros? (rename to inline-tag?)
- State
   - input vs state
   - no-update?
- Styles
- Layouts & containers
- Events
   - listening
   - emitting
- Async & streaming
- Accessibility
   - :scoped
- Animations
- Importing helpers
- Transformers
- 3rd party DOM libraries
   - onMount
   - key + this.getEl
- Web Components
   - registering with html-elements.json
   - wrapping imperative apis
   - using marko in a web component
- Component communication
   - input down, events up
   - key + this.getComponent
- Component composition
- Peak performance
   - NODE_ENV
   - Profiling/Flame graphs
   - Pass in promises

### Reference
- Compiler
   - API
   - AST Nodes & Builder
   - Compile Context
   - `marko.json`/`marko-tag.json`

- Components
   - structure
      - single-file
      - multi-file
      - split
   - variables
   - attributes
   - properties
   - methods
   - lifecycle

- Render API

jasonmacdonald:
Looks like a good start. One thing I might add is more details about what options that are available through the marko.json and marko-tag.json files. The old docs for Marko 3, if I recall, had lots of info on this, but I find I often need to go into the Marko tag-lib source code to see what I can do with those files. There's a lot of power there, but little explanation on how to use it.

Glad to see Transforms in the list, along with compiler stuff. Those would have been my other suggestions.

StarpTech:
I'm going to use markojs in production but as a beginner, I faced with following topics.

- How to enable SASS in lassojs?
- Instead to promote static sites only (marko-cli) we should also show the benefit of dynamic sites. How looks a basic app with a layout (header, footer, body) ?
- How to prebuild static assets with lassojs?
- When is content delivered to the browser? Even an empty `class` will mark everything as client and server code. This isn't obvious.
- Who is using markojs in production? How much is it used at ebay?
- I faced with many outdated or read-only repositories with signal a bad state of the ecosystem. Many issue has to be cleaned up.
- What are the plans for Marko v5? How can we participate?

Thanks you.

DylanPiercey:
@StarpTech we have plans this year to improve the documentation and have started with some additional documentation recently on transclusion, new syntax and configuration. We will be continuing to improve the documentation throughout the year.

[...]

PR #1663, opened 2021-02-20
Copyedit server-side rendering docs

tigt:
It might also be wise to rewrite the Lasso example to come _after_ the other integrations, if the Lasso roadmap is what I think it is.


tigt:
I got hired to copyedit and [restructure all the Marko docs](https://github.com/marko-js/marko/issues/874), so closing this.

PR #1696, opened 2021-05-19
fix: potential issue if translator is not an explicit dependency

DylanPiercey:
## Description
In some cases depending on how `npm/yarn` decides to hoist `@marko/translator-default`, it can make it such that the compiler cannot find the translator. For the default translator we bring it in as a dep of `marko` itself and so don't require users manually install the translator.

This PR re-exposes the translator from the Marko package which should always be discoverable to avoid the issue with the default translator.

## Checklist:


tigt:
Closing, because this improvement would be redundant with a larger change we want to make:

> @DylanPiercey: I think for this one we should go with the alternative of bundling the compiler which makes us more more resilient to babel's breaking changes anyway.

Issue #1727, opened 2021-08-05
Marko 5 ESM Loading/Registering


adamchal:
### Marko Version
5.15.1

### Details

Trying to render a Marko template to HTML using Node in a module (ESM) context. [Minimal repo](https://github.com/adamchal/marko-5-esm-issue) showcasing the issue.

Using `marko.load()` to load the Marko template
```javascript
import marko from "marko";
(async () => {
  const template = marko.load("./test.marko");
  const rendered = await template.render();
  console.log(rendered.getOutput());
})();
```

### Expected Behavior

HTML output of test.marko logged to the console.

### Actual Behavior

```
WARNING!!
Using `marko/compiler` has been deprecated, please upgrade to the `@marko/compiler` module.
  at node:internal/modules/cjs/loader:1095:14

/marko-5-esm-issue/components/foo.marko:1
div -- foo
       ^^^

SyntaxError: Unexpected identifier
    at Object.compileFunction (node:vm:352:18)
    at wrapSafe (node:internal/modules/cjs/loader:1025:15)
    at Module._compile (node:internal/modules/cjs/loader:1059:27)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1124:10)
    at Module.load (node:internal/modules/cjs/loader:975:32)
    at Function.Module._load (node:internal/modules/cjs/loader:816:12)
    at Module.require (node:internal/modules/cjs/loader:999:19)
    at require (node:internal/modules/cjs/helpers:93:18)
    at Object.<anonymous> (/marko-5-esm-issue/test.marko:9:36)
    at Module._compile (node:internal/modules/cjs/loader:1095:14)
```

### Possible Fix

However, if you `marko.load` the component _before_ loading the template, it works.
```javascript
import marko from "marko";
(async () => {
  marko.load("./components/foo.marko"); // now it works
  const template = marko.load("./test.marko");
  const rendered = await template.render();
  console.log(rendered.getOutput());
})();
```

But, since components can reference other components, you’d have to load them in order of least dependent to avoid the error.

I have not found clear documentation on how to do this in Marko 5. Previously, in Marko 4, I would `require('marko/node-require').install()` and then `require` the template. This isn’t working for me in Marko 5 in a CommonJS context.

[...]

PR #1728, opened 2021-08-07
Node esm loader support

DylanPiercey:
## Description

Resolves https://github.com/marko-js/marko/issues/1727

Adds support for loading Marko files directly as ESM via a node [ESM Loader hook](https://nodejs.org/api/esm.html#esm_hooks).

This means you can run node like
```terminal
node --experimental-loader @marko/compiler/loader.mjs ./server.js
```

And importing Marko files will work as is.

---

Primarily this PR contains two main changes.
1. Adds a `loader.mjs` file to the compiler package.
2. Updating any `imports` added when compiling a Marko template to have the filename (including extension) to be more native esm friendly.

## Checklist:

### DIFF
--- a/packages/compiler/loader.mjs
+++ b/packages/compiler/loader.mjs
@@ -1,0 +1,34 @@
+import { URL, pathToFileURL, fileURLToPath } from "url";
+import { cwd } from "process";
+import compiler from "@marko/compiler";
+
+const baseURL = pathToFileURL(`${cwd()}/`).href;
+const extensionRegex = /\.marko$/;
+const format = { format: "module" };
+
+export function resolve(specifier, context, defaultResolve) {
+  const { parentURL = baseURL } = context;
+  return extensionRegex.test(specifier)
+    ? {
+        url: new URL(specifier, parentURL).href
+      }
+    : defaultResolve(specifier, context, defaultResolve);
+}
+
+export function getFormat(url, context, defaultGetFormat) {
+  return extensionRegex.test(url)
+    ? format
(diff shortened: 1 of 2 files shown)


packages/compiler/loader.mjs
```diff
@@ -0,0 +1,34 @@
...
+        url: new URL(specifier, parentURL).href
+      }
+    : defaultResolve(specifier, context, defaultResolve);
+}
+
+export function getFormat(url, context, defaultGetFormat) {
+  return extensionRegex.test(url)
+    ? format
+    : defaultGetFormat(url, context, defaultGetFormat);
+}
```
eyelidlessness, line 22: Heads up that the loader hooks API [has changed](https://nodejs.org/docs/latest-v16.12.0/api/esm.html#hooks) in LTS 16.12.0+ along with 17.x, which will presumably provide the initial API for 18.x. The change eliminates `getFormat`, moving its responsibility to the two remaining functions[^1]:

- `resolve`, as a hint
- `load`, which can use that hint but has the final say

You'll likely want to call this in `resolve` (assuming you need to support 14.x), with the fallback deferring to the result from `defaultResolve`.

[^1]: besides `globalPreload`, which is intended for setup not the dynamic runtime

eyelidlessness (commented):
I'm just taking a peek at Marko internals/activity, spotted this PR, and wanted to give a heads up about this API change.

tigt:
Note this will still _probably_ happen because sometimes computers are terrible, but likely with a disclaimer about using it (and the `require()` hook).

PR #1808, opened 2022-06-09
Update `require` to `import` in docs

tigt:
## Description

Mostly a find-and-replace now that `import` is quite mature. I left the Lasso docs alone because that would _not_ be a simple find-and-replace.

## Checklist:

- [x] I have read the **CONTRIBUTING** document and have signed (or will sign) the CLA.
- [x] I have updated/added documentation affected by my changes.
- [ ] ~~I have added tests to cover my changes.~~

PR #1810, opened 2022-06-13
docs: replace CommonJS with ESM

tigt:
Obsoletes #1808 

## Description

- Not updating the example in `10-awesome-marko-features.md` because that example clearly dates itself by mentioning it’s 2017

- Deleted `server-side-rendering.md` because it wasn’t in the navigation (only a couple backlinks even exposed it), and it was redundant with the Rendering page

## Questions

- A couple places (`marko-vs-react.md`, `why-is-marko-fast.md`) show a Marko template’s rendered output. Should I update it to the uglier, longer outputs we have now? Verbatim or rewritten to be easier to follow?

- ~~Do we even want the `server-side-rendering.md#Bootstrapping: Non-Lasso` section anymore? Is it only for people trying to integrate Marko with a new/unsupported bundler?~~ Obviated by deleting `server-side-rendering.md`

## Checklist:

- ~~[ ] I have added tests to cover my changes.~~


packages/marko/docs/marko-vs-react.md
```diff
@@ -617,20 +617,22 @@ style.less {
 
 ### API
 
-Marko compiles component to JavaScript modules that export an API for rendering
-the component as shown below:
+Marko compiles components to JavaScript modules with exported API for rendering,
```
DylanPiercey, line 620: I don't think this is grammatically correct

packages/marko/docs/marko-vs-react.md
```diff
@@ -617,20 +617,22 @@ style.less {
...
+as shown below:
 
 ```js
-require("./components/greeting")
+import greetingComponent from "./components/greeting.marko";
```
DylanPiercey, line 624: Maybe just condense to the following?

```js
import Greeting
```

tigt: Just here, or do you want to move to `PascalCase` for `.marko` imports across the docs from now on?

packages/marko/docs/marko-vs-react.md
tigt: How about the new version?

packages/marko/docs/marko-vs-react.md
DylanPiercey: In think either `Template` or the component name in pascal case throughout the docs would be good.

tigt: I’ll change the instance here to keep this ticket scoped, but we should probably follow up with @mlrawlings to get a Project entry for that particular consistency across the rest of the docs

Issue #1835, opened 2022-07-29
<await client-reorder> not compatible with tight CSP rules.


vwong:
### Description

`<await() client-reorder/>` is incompatible with tight Content-Security-Policies, specifically ones that block `unsafe-inline`.

### Why

I want to be able to lock down my CSP, but then I can't use `client-reorder`.

### Possible Implementation & Open Questions

With client-reordering, the replacement is streamed as a div like the following, (and then subsequently put into the right spot with JS).

```
<div id="af0" style="display:none"> ... </div>
```

An alternate way of representing this, without breaking CSP is:
```
<style nonce="abc123">
  #af0 { display: none }
</style>
<div id="af0"> ... </div>
```

### Is this something you're interested in working on?

I wish I could, but don't know enough about it.

DylanPiercey:
@vwong this is a good point, we should address it and I like your solve there. If you're interested in taking a shot at it, the relevant code is here: https://github.com/marko-js/marko/blob/main/packages/marko/src/core-tags/core/await/reorderer-renderer.js#L60

vwong:
Hmm. Doesn't look hard at all. I'll create a PR.

PR #1836, opened 2022-07-29
feat: apply cspNonce when client reordering

vwong:
## Description

When client reordering in a presence of a tight Content Security Policy, we should apply styles with a nonce.

Fixes https://github.com/marko-js/marko/issues/1835

## Checklist:


packages/marko/src/core-tags/core/await/reorderer-renderer.js
```diff
@@ -54,13 +57,30 @@ module.exports = function (input, out) {
...
+                result.toString() +
+                "</div>"
+            );
+          } else {
+            asyncOut.write(
+              '<div id="af' +
+                awaitInfo.id +
+                '" style="display:none">' +
+                result.toString() +
+                "</div>"
```
vwong, line 81: I opted to keep the original version, as this results in a shorter HTML if you don't need to worry about CSP. Happy to amend it so that it is consistent across both versions.

DylanPiercey:
Thanks for this @vwong, looks great!

DylanPiercey (approved):

packages/marko/src/core-tags/core/await/reorderer-renderer.js
DylanPiercey: Seems good to me 👍

PR #2076, opened 2024-01-09
Upgrade all deps and configs

DylanPiercey:


nspector:
@DylanPiercey The change from `strip-json-comments` to `@luxass/strip-json-comments` changed your min system requirements for node js as the engine required in `@luxass/strip-json-comments` is `>=18`

DylanPiercey:
@nspector I'll take a look and probably drop back to the older module here.
We run our CI on current node lts which passed for node 16, but I see the engines warnings there now.
https://github.com/marko-js/marko/actions/runs/7467209001/job/20320224037

nspector:
Thank you!

DylanPiercey:
@nspector I kind of forgot that node 16 is actually not LTS. We only support LTS versions of node officially so I don't think it's worth worrying about this. If necessary you can either pin a lower version of Marko, or use `--ignore-engines` when installing since I'm pretty sure this module still works in node 16 anyway.

PR #2186, opened 2024-03-26
docs: Fixing NativeTags demo (interface, not namespace)

patricknelson:
## Description

Minor fix for docs at: https://markojs.com/docs/typescript/#registering-a-new-native-tag-eg-for-custom-elements

Noticed it just repeated `namespace` instead of defining `NativeTags` as an `interface`; this fixes that. I assume it should align with structure of the `NativeTags` already defined in [`packages/marko/tags-html.d.t` ](https://github.com/marko-js/marko/blob/main/packages/marko/tags-html.d.ts#L13).

## Checklist:

### DIFF
--- a/packages/marko/docs/typescript.md
+++ b/packages/marko/docs/typescript.md
@@ -228,7 +228,7 @@ interface MyCustomElementAttributes {
 
 declare global {
   namespace Marko {
-    namespace NativeTags {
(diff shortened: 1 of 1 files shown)


DylanPiercey:
Good catch, thanks!

PR #2201, opened 2024-04-11
perf: improve html body content escaping

DylanPiercey:
## Description

After some recent benchmarking it turns out the current escape utilities for html body content are no longer faster than their regexp counter parts.

This PR updates the implementations to use regexps for escaping instead of manual iteration.

Note: In most cases doing a `.test` call before doing a `.replace` is faster, even though it's technically a bit slower for the cases where we actually escape something.

## Checklist:

### DIFF
--- a/.changeset/serious-phones-burn.md
+++ b/.changeset/serious-phones-burn.md
@@ -1,0 +1,5 @@
+---
+"marko": patch
+---
+
+Improve the html content escape helpers. The existing implementations no longer maintain an edge in newer versions of node and the regexp based versions are faster.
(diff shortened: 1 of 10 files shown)


src-code:
@DylanPiercey This upgrade seems to be causing an change in cases where `<script>` nodes in templates reference undefined values , eg
```marko
$ const myVariable = undefined;
<script type="text/javascript">
  var x = { foo: ${myVariable}, bar: true };
</script>
```
used to render to:
```html
<script type="text/javascript">
  var x = { foo: undefined, bar: true };
</script>
```
but now renders as:
```html
<script type="text/javascript">
  var x = { foo: , bar: true };
</script> 
```
Which is a syntax error.

DylanPiercey:
Shoot. I was normalizing the escape script and style helpers to be more like escape xml (which omits nullish values).

I will revert that aspect of the change since it's a regression. In marko 6 though I think we'll have it normalize the same so I recommend avoiding passing undefined as the interpolated value and instead cast undefined to a string your self.

DylanPiercey:
@src-code this is reverted in #2204 which I'll release shortly.

PR #2209, opened 2024-04-15
New core tags: `<do>`, `<log>`, `<debug>`

LuLaValva:
- `<do>`: Run code in the render phase
- `<log>`: Log to the console
- `<debug>`: Make a debug statement


svallory:
Quick question: will Marko allow me to override these tags if I want to?

LuLaValva:
> Will Marko allow me to override these tags if I want to?

Probably not.

I don't think we've 100% settled on the order that tags are discovered yet, but iirc this is what we are currently thinking:
```
Native tag > Core Tag > Local Variable > Custom Tag
```

DylanPiercey:
Note @svallory these tags only apply to marko 6 fyi. I think we'll likely maintain a similar behavior where you can override built-in tags but I think the verdicts still out before we hit beta.

svallory:
> Note @svallory these tags only apply to marko 6 fyi. I think we'll likely maintain a similar behavior where you can override built-in tags but I think the verdicts still out before we hit beta.

Yeah, I knew that was Marko 6 only.

The reason I asked is because I think those tag names may be going a bit too much into user land. Or at least something belonging to a standard library that is opt-in.

I can definitely see myself wanting to write a <log>, or <debug> tag that uses a specific library under the hood

PR #2230, opened 2024-05-16
Update webpack.md to fix example config

ericbiewener:
## Description

The provided example does not work.

## Checklist:

### DIFF
--- a/packages/marko/docs/webpack.md
+++ b/packages/marko/docs/webpack.md
@@ -60,7 +60,7 @@ export default [
 ```javascript
 import MarkoPlugin from "@marko/webpack/plugin";
 
-const markoPlugin = new MarkoPlugin();
+const markoPlugin = new MarkoPlugin.default();
 
 export default [
   {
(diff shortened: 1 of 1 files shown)


packages/marko/docs/webpack.md
```diff
@@ -76,17 +76,20 @@ export default [
...
-        loader: "@marko/webpack/loader",
-      },
-      // If using `style` blocks with Marko you must use an appropriate loader
-      {
-        test: /\.css$/,
-        use: ["style-loader", "css-loader"],
-      },
-    ],
+    module: {
+      entry: "./browser.js",
```
ericbiewener, line 80: I don't know if this is right. But omitting it leads to the error:

```
TypeError: Cannot read properties of undefined (reading 'markoEntriesPending')
```

DylanPiercey: The marko server plugin should populate the entry for the browser, so if that isn't working then this would be a bug.

PR #2692, opened 2025-06-04
fix: check for errors in ServerRendered.toString

paperclover:
## Description

without this check, a synchronous render with an error will return all of the text up to the error, which is likely to be invalid HTML.

## Checklist:

### DIFF
--- a/packages/runtime-tags/src/html/template.ts
+++ b/packages/runtime-tags/src/html/template.ts
@@ -284,7 +284,9 @@ class ServerRendered implements RenderedTemplate {
     const head = this.#head;
     this.#head = null;
     if (!head) throw new Error("Cannot read from a consumed render result");
-    if (!head.boundary.done) throw new Error("Cannot fork in sync mode");
+    const { boundary } = head;
+    if (!boundary.done) throw new Error("Cannot fork in sync mode");
+    if (boundary.signal.aborted) throw boundary.signal.reason;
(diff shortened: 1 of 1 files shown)


DylanPiercey:
Great catch (pun intended)! Appreciate the PR 😃

paperclover:
i didn't look too much at testing. if someone can point to how i'd write a test for this part of the api (i found only what looked ike snapshot-based tests), i can add tests for this.

some unrelated feedback related to this function, i find the error message "Cannot fork in sync mode" very confusing at first, until you dig more internals to learn that <await> is implemented by a function named `fork`. not sure if that's the only way to fork, but it might help users consuming the template api if it said `Cannot consume asynchronous render with 'toString'`. i can add that here since it's the adjacent line, or do in another patch, or not at all. up to the maintainers.

DylanPiercey:
@paperclover the vast majority of our tests are in snapshots (we basically convert all rendering / errors and test behavior to an md file and snapshot that). However the current Marko 6 suite isn't really setup for testing the sync rendering api specifically since all the snapshots use the streaming apis. I think it'd need a bespoke test.

We should probably add a bespoke test for all of the top level render apis (ssr and csr) but I think we could hold off on that for now and get your patch in.

I agree on the error message, if you want to include the change as you suggested that sounds great to me, otherwise lets merge!

PR #2776, opened 2025-08-24
Improve types for properties of `<style>` as a Tag Variable

### LINKED ISSUE #2775: `<style>` as Tag Variable lacks types for class properties
### Details

As described in the docs for [`<style>`](https://next.markojs.com/docs/reference/core-tag#style):

> If the `<style>` tag has a [Tag Variable](https://next.markojs.com/docs/reference/language#tag-variables), it leverages [CSS Modules](https://github.com/css-modules/css-modules) to expose its classes as an object.

In my usage, I've found this _works at build/runtime_, but I see TypeScript type errors (at least in VSCode) when accessing those properties. For instance, in this snippet from the documented example:

```marko
<style/styles>
  .foo {
    border: 1px solid red;
  }
</style>

<div class=styles.foo/>
//                ^ Property 'foo' does not exist on type '() => HTMLStyleElement'
```

### Expected Behavior

**Ideally:** since `<style>` as a Tag Variable is compiled as a CSS module, it would be very nice if the specific exports were exposed in generated types.

**Pragmatically:** it makes sense that the `<style>` Tag Variable's type could be expressed as `Record<string, string>` (perhaps as an intersection with `() => HTMLStyleElement`?).

### Actual Behavior

As above, accessing properties on the `<style>`'s Tag Variable produces TypeScript type errors.

(Well, to be pedantic… since the variable is typed as a function, accessing a function's properties _does not_ produce type-level errors. I haven't checked the runtime behavior for that, but I could imagine…!)

### Possible Fix

I'll open a PR shortly!

<details><summary>Additional Info</summary>

### Your Environment

- Environment name and version (e.g. Chrome 39, node.js 5.4): N/A. More applicable: I observe the issue in VSCode
- Operating System and version (desktop or mobile): N/A
- Link to your project: I actually don't have one yet! I just started playing with Marko this morning. 😅 But I see the same issue in this repo, i.e. [this fixture](https://github.com/marko-js/marko/blob/34b88ac5642cf7ec85ce69250df39b0b7458e44f/packages/runtime-tags/src/__tests__/fixtures/style-tag-modules-destructured/template.marko#L1).

### Steps to Reproduce

1. Create a Marko template with `<style>` as a Tag Variable (or use the fixture linked above)
2. Open the template in an editor configured to check TypeScript types in Marko

### Stack Trace

N/A

</details>

eyelidlessness:
Fixes #2775, eliminating TypeScript type errors when accessing properties associated with CSS-defined classes.

## Description

As described in #2775, this is an attempt to produce the **pragmatically** expected behavior. I.e. given the same example snippet from the docs/inlined there:

```marko
<style/styles>
  /*   ^ [1] */
  .foo {
    border: 1px solid red;
  }
</style>

<div class=styles.foo/>
/*                ^ [2] */
```

Before:

1. `() => HTMLStyleElement`
2. **Error:** Property 'foo' does not exist on type '() => HTMLStyleElement'

After:

1. `(() => HTMLStyleElement) & Record<string, string>`
2. **Type:** `string`

I’m reasonably confident that this **works** (i.e. I can observe that it addresses the issue I want to address.) Some things I'm less confident about:

- ~~I have some doubts about whether this is the right way to make type-level changes in the project structure _generally_. I originally asked about this on Discord, but decided to just go ahead and open a PR so there's a more concrete place for feedback.~~ @DylanPiercey I think the matching changes should be right based on your Discord response, let me know if I'm mistaken!

- I'm not sure if this approach and the resulting type is actually expected/right! As mentioned in #2775, I'm not sure if the type actually does match the runtime. I'm happy to check, but I'm running out of energy for dev stuff today and wanted to get at least this much open for feedback before I wrap up.

- I'm not thrilled with the more global change: introducing `& any` to the corresponding types for unspecified tags. I'd be happy to address this (e.g. with a conditional type or similar mechanism), but I didn't want to get too bogged down in that until the previous point is addressed.

## Checklist:

Re: documentation, I'm open to direction!

Re: tests, as I've been writing this up @DylanPiercey mentioned on Discord type-level testing in https://github.com/marko-js/language-server. I'd be happy to look into that, possibly tomorrow, if it would be helpful!

### DIFF
--- a/packages/runtime-class/index.d.ts
+++ b/packages/runtime-class/index.d.ts
@@ -352,12 +352,15 @@ declare global {
     export interface NativeTag<
       Input extends Record<string, any>,
       Return extends Element,
+      Value = never,
     > {
       input: Input;
-      return: { value: () => Return };
+      return: {
+        value: (() => Return) & Value;
+      };
     }
     export interface NativeTags {
-      [name: string]: NativeTag<Record<string, any>, Element>;
+      [name: string]: NativeTag<Record<string, any>, Element, any>;
     }
 
     export type Input<Name> = 0 extends 1 & Name
--- a/packages/runtime-class/tags-html.d.ts
+++ b/packages/runtime-class/tags-html.d.ts
@@ -101,7 +101,11 @@ declare global {
       source: NativeTag<Marko.HTML.Source, HTMLSourceElement>;
       span: NativeTag<Marko.HTML.Span, HTMLSpanElement>;
       strong: NativeTag<Marko.HTML.Strong, HTMLElement>;
-      style: NativeTag<Marko.HTML.Style, HTMLStyleElement>;
+      style: NativeTag<
+        Marko.HTML.Style,
+        HTMLStyleElement,
+        Record<string, string>
+      >;
       sub: NativeTag<Marko.HTML.Sub, HTMLElement>;
       summary: NativeTag<Marko.HTML.Summary, HTMLElement>;
       sup: NativeTag<Marko.HTML.Sup, HTMLElement>;
--- a/packages/runtime-tags/index.d.ts
+++ b/packages/runtime-tags/index.d.ts
@@ -110,12 +110,15 @@ declare global {
     export interface NativeTag<
       Input extends Record<string, any>,
       Return extends Element,
+      Value = never,
     > {
       input: Input;
-      return: { value: () => Return };
+      return: {
+        value: (() => Return) & Value;
+      };
     }
     export interface NativeTags {
-      [name: string]: NativeTag<Record<string, any>, Element>;
+      [name: string]: NativeTag<Record<string, any>, Element, any>;
     }
 
     export type Input<Name> = 0 extends 1 & Name
--- a/packages/runtime-tags/tags-html.d.ts
+++ b/packages/runtime-tags/tags-html.d.ts
@@ -101,7 +101,11 @@ declare global {
       source: NativeTag<Marko.HTML.Source, HTMLSourceElement>;
       span: NativeTag<Marko.HTML.Span, HTMLSpanElement>;
       strong: NativeTag<Marko.HTML.Strong, HTMLElement>;
-      style: NativeTag<Marko.HTML.Style, HTMLStyleElement>;
+      style: NativeTag<
+        Marko.HTML.Style,
+        HTMLStyleElement,
+        Record<string, string>
+      >;
       sub: NativeTag<Marko.HTML.Sub, HTMLElement>;
       summary: NativeTag<Marko.HTML.Summary, HTMLElement>;
       sup: NativeTag<Marko.HTML.Sup, HTMLElement>;


packages/runtime-class/index.d.ts
```diff
@@ -352,12 +352,15 @@ declare global {
     export interface NativeTag<
       Input extends Record<string, any>,
       Return extends Element,
+      Value = never,
     > {
       input: Input;
-      return: { value: () => Return };
+      return: {
+        value: (() => Return) & Value;
```
eyelidlessness, line 359: So I came back and checked out the actual runtime behavior and it appears that `<style/foo>` produces a [module namespace object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import#module_namespace_object). In hindsight, this isn't surprising at all. That's exactly what the docs say:

> If the <style> tag has a [Tag Variable](https://next.markojs.com/docs/reference/language#tag-variables), it leverages [CSS Modules](https://github.com/css-modules/css-modules) to expose its classes as an object.

Which suggests that the following should be mutually exclusive:

```marko
<style/> /* () => HTMLStyleElement */
<style/binding/> /* Record<string, string> */
```

And even that isn't _quite right_. The runtime representation of CSS modules may be tooling-dependent. I'm struggling to find this now in the docs, but unless I'm mistaken, I remember reading that this is deferred to the bundler. In any case, what I see with my (mostly) out-of-the-box project (Marko 6 + Marko Run + Vite) is an interface more like:

```ts
interface CSSModule {
  [key: string]: string;
  default: Record<string, string>;
}
```

(That's not actually a valid type definition, but hopefully it gets the point across.)

Anyway... all of this is a long way around the barn to say:

1. I'm even more doubtful that _these types_ are right.
2. I'm actually kind of doubtful that static type definitions are even the right place to solve this!

On Discord, @DylanPiercey also mentioned interest in special casing CSS module typegen in the Marko language server. That sounds like it would produce more appealing types _in the first place_... it would certainly be better than this change in its current state.

I'll defer to the Marko team on whether this PR is worth pursuing any further. But it does feel pretty iffy to me, and I won't mind at all if y'all close it for now in lieu of a more robust/fitting solution.

eyelidlessness:
Another observation: before I explored the solution in this PR, I had originally tried just augmenting the types locally. I found it impossible to define a more targeted augmentation because it would conflict with the underlying types, and the language server just disregarded my augmentation. That said, today I happened to discover a hacky way to do pretty much what I tried the first time around.

If I add `src/tags/style.marko` with...

```marko
<return={} as Record<string, string>/>
```

... ~~so long as I don't import that tag~~[^1] this seems to produce the same approximately-desired effect (at least in VSCode).

[^1]: Apparently this isn't an issue either, presumably because `<style>` is special cased as a builtin: it just shows up as unused, rather than shadowing the builtin as I'd expected.

As a workaround, I think this is sufficient for what I'm trying to achieve right now. It may also be worthwhile if anyone else is looking for a quick way to do away with this particular case of red squigglies in their editor. And maybe that also adds more weight to the approach in this PR not being the best solution for now?

DylanPiercey:
This is now implemented in latest marko-vscode and @marko/type-check with correct types 🎉

Issue #2721, opened 2025-06-23
Cannot pass in input variables into the <await> tag


agliga:
### Details
When using await tag I cannot pass in `input` into a component inside await tag.

### Expected Behavior
I should be able to pass in input from the parent. If I remove the input reference, the code renders fine.

### Actual Behavior

Getting: `An error has occurredTypeError: Cannot read properties of undefined (reading 'attributes')` (this is logged from the `catch` part of the try/catch.

### Your Environment
https://github.com/eBay/evo-web/blob/marko6-3d-viewer/packages/evo-marko/src/tags/evo-3d-viewer/index.marko

DylanPiercey:
The `<await>` here https://github.com/eBay/evo-web/blob/marko6-3d-viewer/packages/evo-marko/src/tags/evo-3d-viewer/index.marko#L58 is going to run on the server where it doesn't make any sense.

In this case you most likely want to kick off the load in the browser, so the code would be more like:

```
<let/loading=undefined>
<script>loading = modelViewerLoad()</script>

// later ...

        <await=loading>
            <${`model-viewer`}
                camera-controls
                touch-action="pan-y"
                src=src
                ...htmlInput
            />
        </await>
```

Could you give that a try?

PR #3027, opened 2026-01-01
fix issue 2721

taimoorzulfiqar:
## Description

Description of Changes
Added element existence checks in _attrs and _attrs_partial to prevent accessing attributes when the element is undefined.

Why This Change Is Required
Fixes issue #2721: components inside <await> tags that use input attributes can crash with TypeError: Cannot read properties of undefined (reading 'attributes') when the element doesn't exist yet. The guards return early if the element is undefined, preventing the crash while preserving correct behavior once the element is created.

link to issue - https://github.com/marko-js/marko/issues/2721

## Checklist:

- [ x] I have read the **CONTRIBUTING** document and have signed (or will sign) the CLA.
- [ x] I have updated/added documentation affected by my changes.
- [ x] I have added tests to cover my changes.

### DIFF
--- a/packages/runtime-tags/src/dom/dom.ts
+++ b/packages/runtime-tags/src/dom/dom.ts
@@ -124,6 +124,10 @@ export function _attrs(
   nextAttrs: Record<string, unknown>,
 ) {
   const el = scope[nodeAccessor] as Element;
+  // Guard against undefined element (can happen in await tags before element is created)
+  if (!el) {
+    return;
+  }
   for (let i = el.attributes.length; i--; ) {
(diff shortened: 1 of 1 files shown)


DylanPiercey:
@taimoorzulfiqar this issue is not the correct fix, basically if the element is undefined it means something else has gone wrong.

There have been a number of `<await>` related issues that have already been addressed and so it is possible #2721 is already fixed also. If you'd like to help, adding a simplified reproduction to the test suite, or sharing a link to a reproduction in [the playground](https://markojs.com/playground) would be great!

PR #3092, opened 2026-02-21
fix: allow parenthesized bind values

LuLaValva:
Sometimes prettier wraps values in parentheses, like `<let/x:=(foo.bar.baz)>`. Until this PR, that led to a compile error!


DylanPiercey:
Superseded by #3097

PR #3183, opened 2026-05-11
feat: optimise serialisation of static content

### LINKED ISSUE #3178: Overserialisation in <for><if>
### Marko Version: 6.0.167

When used like:
```
<for|{ content }| of=input.section>
  <if=content>
    <${content}/>
  </if>
</for>
```

The compiler generates unnecessary JS bundle.

Minimal reproduction: https://github.com/vwong/marko6-over-serialisation

vwong:
## Description

When a child template has no client state, it does not need to be serialised. In some cases, this means that the parent of this child template does not need to be serialised either, even if the parent on its own uses reactive constructs.

This PR adds additional reasons for serialisation to reduce the amount of child templates being serialised.

Fixes https://github.com/marko-js/marko/issues/3178

## Checklist:


packages/runtime-tags/src/translator/util/sections.ts
```diff
@@ -313,14 +313,25 @@ export function getNodeContentType(
...
       downstreamBinding.binding,
       downstreamBinding.properties,
     );
+    if (!downstreamReasons) return false;
+    if (
+      downstreamReasons !== true &&
+      !section.serializeReason &&
+      !section.serializeReasons.size &&
+      !section.parent?.serializeReason &&
+      !section.parent?.serializeReasons.size
```
DylanPiercey, line 330: @vwong I need to think through more if this is safe, however I think at least it would need to recursively check parent sections.

vwong: Good point. Do you want me to try to make the change, or are you taking over?

DylanPiercey: I'm seeing if I can come up with a failing fixture but so far haven't been able to, possible I'm missing something and it's fine?

vwong: 🤔 I'll think about this too (occupied atm). At the very least, this'll optimise shallow cases, even if it [may or may not] handle deeper ones.

DylanPiercey: My main concern is if it breaks deeper ones (doesnt resume content when it should)

PR #3184, opened 2026-05-11
feat: optimise input.content serialisation

vwong:
## Description

In many cases, we want to pass static content from grandparent to parent to child, even if the intermediate parent may have event handlers (thus state-ful). This PR introduces an optimisation that avoids serialising code to generate the otherwise static content..

## Checklist:


vwong:
I'm terrible at finding names, let me know if you want to change the name of the fixture to match existing conventions.

PR #3185, opened 2026-05-13
fix(runtime-tags): skip runEffects for already-destroyed branch scopes

vwong:
## Description

Guard the base `runEffects` loop with
`!scope[ClosestBranch]?.[Destroyed]`, mirroring the identical guard already present in `runRenders` and the `_enable_catch`-enhanced path.

Without the guard, if a branch is destroyed during the same `runRenders()` pass that queued its setup effects, those effects still fire in `runEffects()` on a stale scope where DOM node refs are `undefined`, producing a TypeError at runtime.

Adds a regression test suite (`queue-destroyed-branch.test.ts`) that exercises the four key cases: destroyed-branch effect is skipped, active-branch effect runs, top-level (no-branch) effect runs, and a mixed batch skips only the destroyed-branch effects.

## Checklist:

### DIFF
--- a/.changeset/cozy-ways-nail.md
+++ b/.changeset/cozy-ways-nail.md
@@ -1,0 +1,6 @@
+---
+"@marko/runtime-tags": patch
+"marko": patch
+---
+
+Fix a race condition whereby effects can still fire in stale scopes
--- a/.sizes.json
+++ b/.sizes.json
@@ -7,8 +7,8 @@
     {
       "name": "*",
       "total": {
-        "min": 21171,
-        "brotli": 7956
+        "min": 21193,
+        "brotli": 7940
       }
     },
     {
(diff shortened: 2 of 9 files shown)


DylanPiercey:
@vwong thanks for the pr but I'm not sure this is the right fix.

Basically `runEffects` _should_ only be called immediately after executing render code, and it should not be possible for an effect to be queued from a render that has it's scope destroyed in the same task.

I'm also concerned by the overhead added here on every effect unnecessarily (in theory unless I'm missing something).
Also I'd rather not add bespoke runtime unit tests for something like this, preferably a fixture based snapshot. Usually we avoid unit tests since the runtime is in constant flux and the main thing we want to codify is the behavior as implemented in the template code.

I'm guessing the race condition may actually be here https://github.com/marko-js/marko/blob/dc68e86479d3834e050d6ba63be17553c288dd5f/packages/runtime-tags/src/dom/control-flow.ts#L97, or somewhere else in the way this works with async...

vwong:
@DylanPiercey Thanks for the feedback, I'll come up with something else

PR #3202, opened 2026-05-29
fix: prevent crash firing lifecycle callbacks on closing browser tab

vwong:
## Description

Is it possible that when a tab is closed mid-load, pagehide/visibilitychange/freeze can fire while Marko is still running its mount lifecycle? Could this explain why I'm occasionally seeing crash reports of undefined element references? I've not been able to reproduce locally because of the extreme timing of such an event.

Of course, it doesn't really matter that there's a crash because users won't see it anyway, except for the crash reports in the bug tracker (I'm using BugSnag). Worth fixing?

This PR adds a test fixture demonstrating the problem, and the associated fix.

It does increase the bundle size somewhat though...

## Checklist:


DylanPiercey:
@vwong I don't think hooking into page events should be needed and this means theres a bug elsewhere. Theres been some similar bugs fixed recently so hopefully whatever you were seeing is covered also 🤞

PR #3239, opened 2026-06-20
fix(runtime-tags): quote attribute values containing `=` or backtick

DylanPiercey:
## Problem

`needsQuotedAttr` (`src/html/attrs.ts`) decides whether an attribute value must be wrapped in quotes. Its character class `/["'>\s]|&#?\w+;|\/$/g` omits `=` and the backtick, both of which are forbidden in an unquoted HTML attribute value (and a value may not even *begin* with `=`).

As a result values such as `=x` or a query-string URL like `/p?a=1&b=2` were emitted unquoted:

```html
<a href=/p?a=1&b=2>link</a>
<div data-x==x`y></div>
```

which is non-conforming markup (parser-/tooling-dependent and fragile).

## Fix

Add `=` and the backtick to the must-quote character class. The existing quote-style selection and escaping already handle the rest, so matching values are now emitted quoted:

```html
<a href="/p?a=1&b=2">link</a>
<div data-x="=x`y"></div>
```

## Test

New fixture `attr-quote-equals-backtick` renders attributes whose values contain `=` and a backtick; the `writes.html` snapshot goes red→green (unquoted → quoted) with this change. No existing snapshots were affected (none currently emit an unquoted `=`/backtick value).

---
_Generated by [Claude Code](https://claude.ai/code/session_01CTgmDYu49wQNwQCxRicL3d)_


DylanPiercey:
Closing as a false positive — not a real bug.

Per the HTML tokenizer's *attribute-value-unquoted state*, `=` and `` ` `` (and `"`, `'`, `<`) are flagged as a parse error, but the error recovery is to **treat them as the "anything else" case** — i.e. append the character to the value. So every modern browser parses `href=/p?a=1&b=2` to exactly `href` = `/p?a=1&b=2` with no splitting or data loss (the `&b=2` is left literal by design, to preserve legacy query strings). It renders correctly everywhere.

The only thing the change buys is strict-validator conformance, and it carries a real cost: adding `=` to the must-quote set would force-quote an extremely common pattern (query-string URLs that we currently emit unquoted), regressing output size for no functional gain. The backtick rationale is historical (old-IE XSS quirk), not relevant to modern targets.

If validator-clean output is ever desired, an opt-in would be the better framing than always-quote.

---

PR #3240, opened 2026-06-20
fix(runtime-tags): keep scope identity for nullish `<for by>` keys

DylanPiercey:
## Problem

The keyed `<for>` DOM reconciler (`src/dom/control-flow.ts`) builds its old-scope lookup map with a positional fallback for nullish keys:

```js
oldScopes.reduce((map, scope, i) => map.set(scope[LoopKey] ?? i, scope), new Map())
```

…but then looks branches up by the **raw** `by()` result (`.get(key)`). So an item whose `by` returns `null`/`undefined` is stored under its index yet looked up under `undefined`, never matches, and its branch is destroyed and recreated on every update — losing DOM node identity and local component state, with needless churn.

## Fix

Apply the same positional fallback on the lookup side (`key ?? index`, where `index` tracks the item position to mirror the map's `?? i`). This only changes behavior for nullish keys — for any non-nullish key `key ?? index === key`, so normal keyed lists are unaffected.

## Test

New fixture `for-tag-by-nullish-key` has a 3-item list whose middle item's `by` key is `undefined`. Re-rendering an identical list goes red→green: without the fix the middle `<li>` is removed and re-inserted (churn in the Change log); with the fix the update produces no DOM mutations. Verified against the existing reorder/keyed-`for` fixtures (88 passing) for regressions.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CTgmDYu49wQNwQCxRicL3d)_

### DIFF
--- a/packages/runtime-tags/src/__tests__/fixtures/for-tag-by-nullish-key/__snapshots__/dom.bundle.debug.js
+++ b/packages/runtime-tags/src/__tests__/fixtures/for-tag-by-nullish-key/__snapshots__/dom.bundle.debug.js
@@ -1,0 +1,22 @@
+// template.marko
+const $template = "<ul></ul><button>same</button>";
+const $walks = " b b";
+const $for_content__item_id = ($scope, item_id) => _text($scope["#text/0"], item_id ?? "?");
+const $for_content__$params = ($scope, $params2) => $for_content__item($scope, $params2[0]);
+const $for_content__item = ($scope, item) => $for_content__item_id($scope, item?.id);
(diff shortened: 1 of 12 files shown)


DylanPiercey:
Closing as superseded. This took the "make `<for by>` silently tolerate nullish keys" approach, but we decided to handle that bug the opposite way — by emitting a `MARKO_DEBUG` warning that a `by` key must be a string or number (a nullish/object key can't track identity or survive serialization/resume). That shipped in #3246. Keeping this open would conflict with that direction.

---

PR #3280, opened 2026-06-23
perf(runtime-tags): render conditional native attribute values at compile time

DylanPiercey:
Applies the placeholder text-escaping optimization to native attribute values. The `_attr`/`_attr_class`/`_attr_style` helpers are distributive over a conditional, so they're pushed down through the branches of a conditional attribute value and literal branches are serialized at compile time:

| Source | After |
|---|---|
| `title=x ? "active" : "idle"` | `x ? " title=active" : " title=idle"` |
| `class=x ? "on" : "off"` | `x ? " class=on" : " class=off"` |
| `aria-hidden=z ? true : null` | `z ? " aria-hidden" : ""` |
| `lang=x ? "en" : fallback` | `x ? " lang=en" : _attr("lang", fallback)` |

Static branches do zero runtime helper work (and get the optimal unquoted/bare form at build time); mixed branches keep the dynamic side at runtime. HTML-only — the DOM path already sets attributes without escaping. Non-conditional class/style lists (`class=["icon", {active}]`) are unchanged.

Follow-up: extract the static parts of class/style *lists/objects* on the HTML side too (the DOM side already does this), likely via a nested class-accumulator helper.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JZakV7agrS2UMmEGaTGkiq)_


DylanPiercey:
Folded into #3281, which squashes this conditional-attribute optimization together with the logical-expression and static `class` extraction follow-ups into a single commit against `main`.

---

PR #3508, opened 2026-07-21
fix: harden RegExp serialization against an own source/flags breakout

DylanPiercey:
`writeRegExp` read `val.source` / `val.flags` directly off the instance and spliced them into the emitted `/.../` resume literal. Both are `RegExp.prototype` accessor getters, so an **own data property** — `Object.defineProperty(re, "source", { value })` — shadows them and returns arbitrary text, including a `/` that closes the literal and injects expression-position code into the serialized `<script>`.

Repro (before): a `/x/` with an own `source` of `a/,x:((globalThis.__PWNED=1)),y:/b` serialized to `/a/,x:((globalThis.__PWNED=1)),y:/b/`, so evaluating the payload set `globalThis.__PWNED`.

Fix: reconstruct through a trusted `new RegExp(val)`, which per spec reads the engine's internal `[[OriginalSource]]`/`[[OriginalFlags]]` slots and ignores own shadows — so only the real source/flags reach the payload. Normal regexes are unaffected (a fresh `RegExp` from a `RegExp` is faithful, and the existing `</script>` escaping still runs).

Adds a serializer test (verified it fails without the fix, emitting the breakout payload). Realism is low — a shadowed-`source` `RegExp` does not arise from normal data flow or `JSON.parse` — so this is hardening/consistency with the file's other own-property-shadow guards. Full serializer suite passes; `build:sizes` unchanged.


DylanPiercey:
Closing — this defends a non-threat at a real cost. An own `source`/`flags` data property on a RegExp only arises from a deliberate `Object.defineProperty` (assignment no-ops on the getter-only accessor; literals / `new RegExp(str)` / `JSON.parse` never produce it), so there is no attacker-controlled data path — the precondition implies the attacker already has server-side code execution, and no trust boundary is crossed. Meanwhile the fix allocated a fresh `new RegExp(val)` on every serialized regex per render. Not worth it.

PR #3532, opened 2026-07-22
fix: dismiss an ancestor @placeholder when a nested await rejects on the client

DylanPiercey:
In a pure client render of `<try @placeholder> … <try @catch><await=rejecting>`, when the inner await rejected the reject path zeroed the ancestor placeholder's pending counter (`awaitCounter.i = 0`) *without* completing it, so `dismissPlaceholder` never ran — the placeholder stayed shown forever and the `@catch` content rendered only into the detached tree.

The reject now completes the counter via `c()` when the await sits under an ancestor `@placeholder` (whose `c()` only unwinds the placeholder). A bare await branch keeps the force-zero, since its `c()` does stream-node surgery that must not run on reject — and that case is exactly `!tryPlaceholder`, so the two are cleanly distinguished.

Verified against the existing `catch-reject-nested-in-await` fixture: CSR now ends on `caught: ERROR!` (matching SSR) instead of stuck on `loading outer...`. Full suite green (6341), no other render/writes changes; net byte impact +10 min / −12 brotli.


DylanPiercey:
Duplicate of #3407, which is more complete: it also guards the resumed-counter case (`!awaitCounter.m`) and fixes the parallel lazy-tag load-error path in load.ts. Closing in favor of #3407.

PR #3533, opened 2026-07-22
fix: drop a sync-failing <try> body's dead scopes from the resume payload

DylanPiercey:
When a `<try>` body throws synchronously and `@catch` replaces its DOM, every scope the body wrote before failing still shipped in the SSR resume payload as a dead fill (and any scope that serialized a registry reference retained that client implementation). The body renders into a chunk that shares the parent `serializeState`, so its `writeScope` records land in the parent map and were never removed on the sync-error path.

`tryCatch` now snapshots the monotonic scope-id counter before rendering the body and, on sync error, drops every scope written at or above that mark — cheap (one number, zero cost on the success path) and safe because a sync body render is contiguous.

New `try-catch-drops-failed-body-scopes` fixture verifies the failed body's `n` scope is dropped while a sibling `count` scope outside the try still serializes and resumes. Full suite green (6357), no other snapshot changes; server-side only. The async reorder-abort path is left for a follow-up (noted in agent-feedback) — its scopes interleave with siblings, so the id-range trick doesn't transfer.


DylanPiercey:
Closing as won't-fix: a `@catch` firing is rare in practice, and this only addressed the sync path (the async reorder-abort path entangles with the bodyNext machinery for little real-world benefit). Capturing the decision as a code comment instead.

PR #3565, opened 2026-07-23
docs: cover $global sharing, scoped styles, and TypeScript in the cheatsheet

DylanPiercey:
Closes four `unclear.md` documentation gaps, all verified against source:

- **Sharing data across components** — new section: no context API, use `input` or request-scoped `$global`, with the serialization caveat. Only allow-listed keys are serialized (`getFilteredGlobals`), and @marko/run defaults to `params`/`url` only, so client-reactive code reading `$global.data` throws after resume while the SSR HTML looks correct.
- **Scoped `<style/name>`** — the cheatsheet implied global CSS was the only option.
- **TypeScript** — new section: exported `Input` interface, `Marko.Body`, `Marko.AttrTag`, generics, and that `tsc` silently skips `.marko` files so `mtc` is needed.
- **`on-attach`/`on-detach`** — added to the 5-to-6 lifecycle table, noting `<lifecycle onDestroy>` cannot defer removal the way Marko 5 could.

Plus four DON'T-table rows and an `mtc` checklist item in the best-practices skill.


DylanPiercey:
Addressed both CodeRabbit findings.

**`<show>` in the exit-animation note** — valid, and my suggestion was wrong. `translator/core/show.ts` documents that a hidden `<show>` parks its nodes in a *detached document fragment*, so a CSS exit transition can never run. Reworked the note to stage removal through state (mark the item `exiting`, drop it in a `transitionend`/`animationend` handler) and to state plainly that nothing in Marko 6 can delay removal.

**"No context API" wording** — narrowed to "No provider/consumer context API", since the same bullet goes on to cover request-scoped `$global`.

PR #3566, opened 2026-07-23
feat(runtime-tags): serialize Intl formatters and Intl.Locale

DylanPiercey:
`Intl` values reached from browser-updating content aborted SSR, so the canonical `<const/fmt=new Intl.NumberFormat(...)>` crashed. They now serialize: formatters rebuild from `resolvedOptions()` and `Intl.Locale` from `toString()`. No client bundle cost.

`Intl.DateTimeFormat` can render `month`/`weekday` differently after resume in some locales (`ja`, `zh`) because its resolved options don't round-trip; noted in the changeset.

Also keeps the earlier commit's addition to the `Unable to serialize` message, and files agent-feedback for other built-ins the serializer still rejects.


DylanPiercey:
Fixed — trimmed that comment to two lines. The `DateTimeFormat` locale caveat it carried is already stated in the changeset and the PR description, so nothing is lost by dropping it from the code.

PR #3569, opened 2026-07-23
docs: clarify that a derived <const> read in a handler is stale until flush

DylanPiercey:
Inside an event handler a `<const>` reads its pre-handler value (updates batch), while a reassigned `<let>` reads back immediately. Reading the derived value expecting the just-set result mis-branches by one step.

Documented generally in the cheatsheet's state rules (where `<const>` auto-recompute is introduced) and in the 5-to-6 migration mapping. A compiler/dev warning was considered and rejected — proving a read is stale needs aliasing analysis, and a value that recomputes synchronously on read is a false positive — so docs is the resolution and the agent-feedback entry is dropped.


DylanPiercey:
Fixed. The example implied `applyMove()` was an increment-by-one while the fix said `moves + 1`, which only holds for +1 moves. Reworked to `moves = next` → read `moves >= max` (the reassigned `<let>` is current in-handler), so it's consistent for any move and demonstrates the actual point — read the fresh `<let>`, not the stale derived `<const>`.

PR #3571, opened 2026-07-23
feat(runtime-tags): serialize DataView

DylanPiercey:
`DataView` was the only `ArrayBuffer` view the constructor dispatch missed while every `TypedArray` is handled. It writes its buffer through so sibling views share one binding, and keeps byteOffset/byteLength for partial views.

Both it and `writeTypedArray` now reject a view whose buffer cannot be written. Previously `new Int8Array(new SharedArrayBuffer(8), 0, 2)` emitted `new Int8Array(,0,2)` — a syntax error that broke the **entire** resume payload, taking unrelated state with it, rather than just reporting that value unserializable. No tampering required; plain constructors reach it.

Boxed primitives are deliberately not supported; noted at the dispatch site. Server-only, no client bytes.


DylanPiercey:
Ran a codex review. Two real defects fixed, three findings triaged as consistent with existing conventions:

**Fixed** — `new DataView(new SharedArrayBuffer(8), 0, 2)` emitted `new DataView(,0,2)`: `writeProp` fails for an unsupported buffer and writes nothing, leaving a syntax hole that breaks the *entire* resume script, not just that value. Now rejects a non-`ArrayBuffer` buffer before pushing anything.

**Fixed** — `writeBoxed` called the overridable `val.valueOf()` and concatenated the result raw, so an overridden `valueOf` put unescaped text into the payload, and an object merely posing as a wrapper (`Object.create({ constructor: Number, valueOf: () => 7 })`) serialized as `Object(7)`. Now uses the intrinsic `Number/String/Boolean/BigInt.prototype.valueOf`, which brand-checks the wrapper and guarantees the emitted primitive's type; a non-wrapper is unserializable.

**No change** — boxed `-0` → `Object(0)` matches plain `-0` → `0`, which `writeNumber` documents as deliberate. Dropped own properties match every built-in writer (`new Date(0)` drops them too). Resizable-buffer semantics are a pre-existing `writeArrayBuffer`/`writeTypedArray` limitation, inherited rather than introduced.

Filed the same `SharedArrayBuffer` hole in `writeTypedArray` (pre-existing, `new Int8Array(,0,2)`) to `agent-feedback/bugs.md`.

PR #3593, opened 2026-07-25
docs(agent-feedback): record why foreign-content escaping needs runtime namespace tracking

DylanPiercey:
`getRawTextEscapeHelper` picks the escaper purely by tag name, so `<style>`/`<script>` inside `<svg>`/`<math>` get the raw-text escapers even though foreign content parses their children as markup — a bare `<` in an interpolation becomes a live element.

The compile-time fix does not work: a static ancestor walk only sees one template's lexical structure, so a `<html-style>` reaching foreign content through a component boundary still picks the raw-text escaper. Nor is any single escaper safe in both contexts, since `<`→`&lt;` corrupts raw-text bodies.

This updates the entry to record that, and that the chunk context (`withContext`, as used for controlled `<select>`) is the viable channel at effort:high, since it breaks flat-string concatenation for all SVG. Docs only.

### DIFF
--- a/agent-feedback/bugs.md
+++ b/agent-feedback/bugs.md
@@ -1570,34 +1570,50 @@ without the click resolves to `<span>VALUE</span>`.
 
 ## Escape `<html-script>`/`<html-style>` interpolations as markup inside `<svg>`/`<math>`, where they are not raw text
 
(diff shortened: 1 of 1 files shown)


DylanPiercey:
Closing — leaving the entry as-is rather than landing a docs-only change for it.

PR #3604, opened 2026-07-26
fix(runtime-tags): read scope-flush slot deltas as signed

DylanPiercey:
`writeScopesRoot` addresses each scope slot by a delta from a running cursor, but only emitted a delta when the next scope id was *higher* than the cursor. A scope flushed after one with a larger id emitted no delta at all, so its props landed in whichever slot the cursor happened to hold — silently attributed to another scope.

The delta is now written signed, with two tests covering it (a later-then-earlier pair, and a scope that flushes twice in one payload). The client already accumulates signed deltas, so nothing changed there, and no payload the current writer produces is affected: it flushes in ascending id order, which makes the negative branch unreachable today.

### DIFF
--- a/.changeset/scope-flush-signed-slot-delta.md
+++ b/.changeset/scope-flush-signed-slot-delta.md
@@ -1,0 +1,5 @@
+---
+"@marko/runtime-tags": patch
+---
+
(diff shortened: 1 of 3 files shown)


DylanPiercey:
Superseded by #3605 (same commit, correctly named branch).

PR #3607, opened 2026-07-26
perf(runtime-tags): resume single-node branches in linear time

DylanPiercey:
The single-node branch-end walk skips a render's own markers to find each branch's start node, and tested membership with `visits.indexOf` — `visits` being every marker comment in the render. Marker elision emits one marker per branch for `<for>`/`<if>` bodies reduced to a single node, so resuming an N-row list cost ~N² identity comparisons, and the loop's terminating test (the first non-marker sibling) always scanned the whole array.

Simulating the walk over a `row, marker, row, marker, …` chain:

| N | `indexOf` | `Set` |
|---|---|---|
| 100 | 10,000 cmp / 0.25 ms | 100 cmp / 0.06 ms |
| 1,000 | 1,000,000 cmp / 1.50 ms | 1,000 cmp / 0.11 ms |
| 10,000 | 100,000,000 cmp / 111.37 ms | 10,000 cmp / 0.59 ms |

The set is built on first use per pass, so a page with no single-node branch resume never allocates one, and it is dropped at the start of each pass because streamed content appends markers between them. Cost: **+52 B raw / +18 B brotli on the dom runtime**, nothing on html — that's the per-fixture size churn in the diff (the resume runtime is inlined into fixture entry chunks).

Behavior is unchanged, so `for-single-node-streamed-branches` pins the shape rather than a regression: single-node branch resume across two streamed boundaries, which instrumenting the pass counter shows is three resume passes with two set builds. Nothing else in the suite covers that.


DylanPiercey:
Closing: the win does not justify the floor cost.

Measured per-comparison cost is 1.11 ns (111 ms / 1e8 comparisons), so at realistic sizes the quadratic scan costs 0.003 ms at 48 single-node branches, 0.011 ms at 100, and 0.069 ms at 250. It only reaches ~1.1 ms at 1,000 branches in a single loop. The set costs +18 B brotli in the dom runtime for every page that resumes a branch.

Recording the crossover in agent-feedback/perf.md and a note at the call site so it is not re-mined naively.

PR #3619, opened 2026-07-26
fix(runtime-tags): stop retaining tags-to-class compat scopes

DylanPiercey:
`classIdToBranch` is filled by the `SET_SCOPE_REGISTER_ID` resume handler for every resumed scope carrying a class component id, but its only `get`/`delete` live in `compat.render`, which runs for the class-to-tags direction alone. Scopes registered by the tags-to-class direction are therefore inserted once and held for the life of the page, keeping the branch scope alive — and after its first re-render, the attached `___marko5Component` and its `___rootNode` fragment — long after the enclosing branch is destroyed.

The directions are distinguishable at registration: tags-to-class writes `{ m5c, m5i }` while class-to-tags writes `{ m5c }` alone.

No fixture can pin this one — the map is module-private and the difference is retention, which needs a GC observation the harness can't make. The evidence is that all 64 interop tests still pass, so the lookup the map exists for is intact.


DylanPiercey:
Closing: I cannot produce a failure case, so this is a bundle-size increase (+16 B min / +52 B brotli on the interop bundle) resting on a code-reading argument alone.

The retention is real by inspection — only `compat.render` reads and deletes from `classIdToBranch`, and it runs for the class-to-tags direction only — but it is not observable from the suite: the map is module-private and the defect is retention, which needs a GC observation the harness cannot make. An out-of-suite `--expose-gc` + WeakRef probe would prove it, but that is not a test anyone can keep.

Leaving the perf.md entry open with the measured cost and the testability problem recorded, so the next attempt starts from what is now known. Worth revisiting if a stale entry can be shown to be *consumed* — a later class-to-tags render whose component id collides with a retained one would morph into a destroyed branch, which would be a correctness bug with a real test.

PR #3694, opened 2026-07-29
test: attribute SSR payload bytes per fixture

DylanPiercey:
Fixture `sizes.json` only reported a total for the SSR payload, which says nothing about where the bytes go. This splits it into disjoint spans — markup, resume markers, reorder wrappers, inlined runtime, resume payload — that are asserted to tile the payload exactly, so a size change can be traced to what caused it.

Test-infrastructure only; no runtime or generated-output change (`.sizes.json` is unchanged).


DylanPiercey:

PR #3695, opened 2026-07-29
feat(compiler): add a `persisted` config option

DylanPiercey:
Translators need to know when they are compiling a document that survives navigation, so a later server response can address what is already there. This adds the option and threads it through as an ordinary Marko config value.

Analysis is cached and shared across outputs, so `persisted` splits the compile cache — otherwise a persisted and an ordinary compile in one process would reuse each other's analysis.


DylanPiercey:

PR #3696, opened 2026-07-29
feat(runtime-tags): keep persisted documents addressable

DylanPiercey:
Stacked on #3695.

Structure and value are two different questions. With the compiler's `persisted` option, a section's scope, its branch marker, node markers and the placeholder separator are emitted unconditionally — the document stays addressable — while what a render actually serializes is unchanged.

Client output must not move at all, so the fixture harness gains a `persisted` mode that asserts the dom bundle is byte-identical to the ordinary build and snapshots the html/SSR side separately. A fully stateful fixture (`basic-counter`) comes out identical either way; the cost lands only where server structure was previously disposable.

### DIFF
--- a/.changeset/lucky-pears-jam.md
+++ b/.changeset/lucky-pears-jam.md
@@ -1,0 +1,5 @@
+---
+"@marko/runtime-tags": patch
+---
+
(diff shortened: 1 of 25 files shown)


DylanPiercey:

PR #3697, opened 2026-07-29
test(runtime-tags): spread the persisted mode across a dynamism range

DylanPiercey:
Stacked on #3696.

One fixture cannot say what the persisted spine costs, so this opts six more into the persisted mode: inert structure, branches, a keyed loop, a dynamic hop, an async boundary and a catch.

Client bytes are identical in every case. Wire cost tracks how much state the template already had — a fully stateful fixture pays nothing, a structure-only one pays several times its (very small) payload:

| fixture | dom off/on | wire off | wire on |
|---|---|---|---|
| if-tag | 375/375 | 22 | 357 |
| dynamic-tag-name | 375/375 | 310 | 637 |
| basic-inert-collapsible-tree | 2844/2844 | 648 | 781 |
| await-tag | 3145/3145 | 711 | 750 |
| basic-counter | 2604/2604 | 363 | 363 |
| for-by | 7485/7485 | 964 | 964 |

The small-payload ratios are dominated by the ~245 B walker runtime that persisted forces in, not by the spine itself.


DylanPiercey:

PR #3699, opened 2026-07-30
fix(runtime-tags): unsubscribe closure scopes adopted from a server render

DylanPiercey:
`subscribeToScopeSet` only recorded cleanup for a scope it actually added to the owner's set. SSR builds those sets with their subscribers already in them and resume adopts them wholesale, so on the client the membership check was already satisfied and nothing was recorded — destroying such a scope left it in the owner's set for the owner's lifetime.

Keys the record off the record itself so an adopted membership still registers, and drops destroyed scopes as `_closure` encounters them, since a scope adopted from the server may never run that signal at all.

Predates the move off per-scope `AbortController`s, which installed its abort listener inside the same membership check. Stacked on #3698 for that reason; the new unit test fails without this change.


DylanPiercey:
Superseded by #3700, which lands the same fix directly on `main` instead of stacked behind the perf change, and carries the red/green test.

PR #3703, opened 2026-07-30
perf(runtime-tags): hold a lone child branch without a Set

DylanPiercey:
Most branches parent exactly one child, but every one of them allocated a `Set` to hold its children. Hold a lone child directly instead and pay for the `Set` only when a second child arrives; the two shapes are told apart by `$global`, which a scope always carries and a `Set` never does.

Worth ~4.5% on tearing down and rebuilding a large keyed list.

### DIFF
--- a/.changeset/lucky-hounds-clap.md
+++ b/.changeset/lucky-hounds-clap.md
@@ -1,0 +1,5 @@
+---
+"@marko/runtime-tags": patch
+---
+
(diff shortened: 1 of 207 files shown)


DylanPiercey:
Closing: re-benchmarked against current main and the change shows no measurable gain. The ~4.5% it was based on was measured before #3698 landed; that PR removed the branch-teardown cost this was helping with, so the remaining Set allocation isn't material. Not worth +34 brotli.

PR #3709, opened 2026-07-30
test: use mocha's built-in parallel runner

DylanPiercey:
Replaces `scripts/test-parallel.js` (custom worker packing + `MARKO_TEST_SLOT_*` env slicing) with `mocha --parallel`. The only thing mocha couldn't do was split a single huge spec file, so `main.test.ts` becomes `main.ts` exporting `main(shard, total)` plus sixteen three-line `main.shard-N.test.ts` entries; mocha's per-file dispatch handles the rest, and the runtime-class suites no longer need slicing at all (the mocha-autotest patch hunk goes away).

Same 10318 tests, and slightly faster: 38.6s vs 40.9s at 4 jobs (CI's core count), since mocha's dynamic dispatch balances better than the script's hardcoded cost hints. `.mocharc.parallel.cjs` pins `jobs` to `availableParallelism()` because mocha otherwise defaults to cores-1, which measured ~2x slower on a 4-core box. `snap.ts`'s snapshot pruning moves to a root hook — parallel workers reuse modules across files, so a top-level `after` would only have run for each worker's first file.


DylanPiercey:
Wrong approach — mocha's file-granular parallelism needs checked-in shard files, and the fixture import rewrite made the diff unreviewable. Reopening as a smaller change to the existing runner.

PR #3711, opened 2026-07-30
perf(runtime-tags): let a conditional bound its own section

DylanPiercey:
A section whose only content is an `<if>` chain was still emitting boundary markers and an anchor to delimit a range the conditional's own branch already bounds — three comment nodes per item for a `<for>` over a conditional. The section now emits none of them, and the rendered branch *is* the section's range; on each swap the conditional retargets any owner still holding the replaced range. When the chain renders nothing its anchor bounds the section instead, so this covers a chain with no `<else>` as well.

Worth ~12% on rebuilding a large keyed list, and removes every comment marker the conditional produced (42 to 0 in a chat-style fixture, matching a markerless framework's node count exactly). Costs +84 bytes brotli.

### DIFF
--- a/.changeset/olive-cats-invent.md
+++ b/.changeset/olive-cats-invent.md
@@ -1,0 +1,5 @@
+---
+"@marko/runtime-tags": patch
+---
+
+Let an exhaustive conditional that is a section's only content bound the section itself, so no marker or anchor nodes are created for it. A `<for>` over `<if>`/`<else>` now allocates one fewer node per item and renders no comment markers for the conditional.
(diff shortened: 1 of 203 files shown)


DylanPiercey:
Superseded by #3712, which gates the same two `getSectionMeta` lines on a strictly weaker condition (any `if`/`for`/dynamic tag at a section edge, rather than an exhaustive conditional as the whole content), so every shape this elides is elided there too — verified by compiling this PR's own target shape on both branches. #3712 also drops the leftover leading `over(1)` walk step this leaves behind, and extends the same owner-retargeting walk to `<for>`, `<show>`/`<await>`/`<try>`, lazy tags and the resume path.

PR #3745, opened 2026-08-05
docs(runtime-tags): correct and extend the agent-facing reference docs

DylanPiercey:
Clears ten `agent-feedback/unclear.md` entries — documentation an agent or contributor was measurably misled by — and files one new `dx.md` entry.

Corrections: `runtime-tags/AGENTS.md` credited `util/assert.ts` with two helpers it does not export and pointed native-element work at `util/known-tag.ts`, which holds tag input contracts; the accessor-lockstep guidance still named the `accessor[.debug].ts` barrels rather than the nine `.debug` pairs; `compiler/AGENTS.md` claimed every stage merges taglib and translator visitors. Fixture anatomy gained the `diagnostics`/error snapshot kinds and the six missing `TestConfig` keys.

Additions: the four `Scope[AccessorProp.Gen]` states, the `Marko.Global` augmentation caveat on the interface itself, cheat-sheet coverage of `static` and of `<let>`-as-a-formula, and a Marko-5-only note on four compiler options no Marko 6 translator reads. The dynamic `<style>` placement warning now says which elements the values actually reach.

`hot` is deliberately left unannotated — it is an unbuilt Marko 6 feature rather than legacy surface area, so it is recorded in `dx.md` instead.


DylanPiercey:
Reviewed the three CodeRabbit findings; applied one, declining two.

**Applied — two-line JSDoc on `meta`.** The repo convention is real (`AGENTS.md`: comments never exceed two lines) and my block was four. Condensed to one line, keeping `@deprecated` and both behaviours.

**Declining — split the changeset by package.** The stated rule does not exist. `AGENTS.md` says to *name* a workspace package and verify with `changeset status`, which is about using a valid name, not a single one. Multi-package changesets are precedented here — `.changeset/comptime-evaluation.md` names `@marko/runtime-tags` and `@marko/compiler` together — and `pnpm exec changeset status` reports both patch bumps correctly.

**Declining — add the `Marko.Global` guidance to `cheatsheet.md`.** It was there and was deliberately removed: `cheatsheet.md` is a dense LLM syntax reference, and a `$global` typing digression does not earn its place. The caveat now lives on the `Marko.Global` interface in `index.d.ts`, where someone augmenting it will actually be looking, and the `unclear.md` entry was narrowed rather than closed to record that the user-facing half belongs in the website repo.

PR #3753, opened 2026-08-06
chore(runtime-tags): register construct-init renders explicitly (experimental)

DylanPiercey:
`_resume_init` becomes a real helper instead of a `_resume` alias: it registers the closure signal's per-branch `_` render — the only thing constructs consume — so every shell-token id (init or mount effect) resolves to a plain branch-callable fn and `patch-branch.feat` drops its `(fn._ || fn)` shape-sniffing. Retention contract unchanged (fused `@__PURE__` onto the declaration; a shaken signal takes the registration with it and constructs fail closed). Translator untouched; bundles now keep the distinct `_resume_init` name instead of folding to `_resume`. Sizes are a wash.


DylanPiercey:
Dropping this for now per dpiercey — hygiene-only, nothing depends on it.

PR #3755, opened 2026-08-06
feat(runtime-tags): persisted client-owned custom tag instances (experimental)

DylanPiercey:
Per-instance classification for state-fed custom tag attributes (last tier-1 item). An instance whose input carries only client state and constants is client-owned: the parent wraps the child render + `_patch_child` link in `if (!_writes_patches())`, so patches skip the subtree — the absent entry keeps the live child, its state, and its handlers (prototype research confirmed it has no answer here; the absent-key protocol is the only sound mechanism). Everything else fails closed at compile time: state+server mixing across one tag's input, tag variables, body content, and `$global` anywhere in the child's template chain (`trackGlobalReference` stamps `metadata.marko.persistedGlobals`; custom-tag analyze rolls it up the load chain, transitively, dynamic tags included). Classification covers per-attr extras, tag arguments, AND the tag's own merged extra — the last one closes a silent-stale hole for rest-consumed children (whole-`input` reads) that predates this stage.

Grok 4.5 (high) adversarial review: round 1 HOLD on exactly that rest-path hole (verified by compiled probes); fixed + pinned by `persisted-child-rest-attr`/`-rest-mixed`, its full scorecard otherwise clean (scope accounting, $global recursion/dynamic-tag roll-up, bound `:=`, non-persisted purity). Fixtures: `persisted-child-state-attr` (child param + child-local state survive server patches; frames carry no child entry), `persisted-child-state-arg`, `persisted-child-mixed-attr`, `persisted-child-global-owned`, plus the two rest shapes. The future per-param withhold channel (to lift the mixed rejection) is recorded in the design note. Non-persisted output unchanged.

### DIFF
--- a/packages/runtime-tags/src/__tests__/fixtures/persisted-child-fn-server/__snapshots__/error-compile-dom.debug.txt
+++ b/packages/runtime-tags/src/__tests__/fixtures/persisted-child-fn-server/__snapshots__/error-compile-dom.debug.txt
@@ -1,0 +1,9 @@
+
+    at packages/runtime-tags/src/__tests__/fixtures/persisted-child-fn-server/template.marko:3:3
+      1 | <let/count=0/>
+      2 | <main>
+    > 3 |   <dump value=count format=(v) => v + input.suffix/>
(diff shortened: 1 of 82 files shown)


DylanPiercey:
Closing — folding this into a larger complete stage rather than landing incrementally.

PR #3861, opened 2026-08-08
fix(runtime-class): resume a class parent's inline function props on a tags child

DylanPiercey:
An inline handler crossing a Class→Tags boundary serialized as a noop, so it only worked after the parent rerendered — and never fired at all under a split parent. The class translator now hoists eligible handlers (those whose free variables are limited to `component`) to a module-scope factory emitted in both builds under a stable id, so each side resumes it through the tags registry; the browser rebuilds the handler lazily against whichever component the resumed scope names. Handlers bind to a dedicated small scope rather than the boundary scope to keep serialization narrow.

Pending hoisted functions are held on the render's `$global` (with a host id per entry) rather than a process-global queue, so a synchronous re-entrant render cannot drain another render's entries and an aborted render cannot leak them into the next drain.


DylanPiercey:
Superseded: these commits are now on `dpiercey-ws-agent-feedback-4tecyk` rebased onto current `main` (see follow-up PR). Closing this branch PR to avoid duplicate review.

DylanPiercey:
Follow-up PR: https://github.com/marko-js/marko/pull/3997

PR #3946, opened 2026-08-13
fix(runtime-tags): report tag variables that form a dependency cycle

DylanPiercey:
Tag variables that depend on each other compiled clean and failed at render — `<const/a=b+1/><const/b=a+1/>` threw `Hoisted values must be functions`, while `<let/a=b/><let/b=a/>` and a custom-tag pair rendered `() => _hoist_read_error()` straight into the page.

Adds an init-order cycle check to `finalizeReferences`, between the loop that records reads and the one that consumes value expressions. Edges are a binding's `upstreamAlias` plus every eager read on its declaring tag's inputs, so it is not tied to `<const>`/`<let>` — a custom tag's var is included because `const {a} = renderChild({ value: b })` needs `b` regardless of whether the child reads it. Reads inside function bodies are `deferred` and excluded, keeping mutual recursion and hoisted helpers legal.

Only a binding hoisted (read before declared) can close a cycle, so the start set is populated in `addBindingGetter` and a template with no hoists pays one `.size` check.

Replaces the codegen-completeness assertion this PR previously carried: that caught only the alias shape, and only because a pure alias falls out of `section.bindings`.


DylanPiercey:
Skipping this one. The check is workable but needs deliberate exclusions (controllable `<let>`, accessor bodies) to avoid rejecting valid templates, which trades away part of the coverage that motivated it.

PR #3976, opened 2026-08-15
docs(runtime-tags): cheat sheet row for an html comment in a tag's attributes

LuLaValva:
An `<!-- … -->` written where an attribute belongs is a mistake coding agents make, and the cheat sheet is what they get pointed at when a compile error is thrown, so it gets a `DON'T` row directing them to `/* … */`.

The parser-level fix for the error itself is marko-js/htmljs-parser#239: `<!--` inside an open tag now reports a targeted `INVALID_HTML_COMMENT` instead of being read as tag type arguments or, after an attribute value, silently consumed as a less-than operator. Marko picks that up through its existing `htmljs-parser` range once it publishes, so nothing is needed in the compiler here.

Also carries an unrelated `agent-feedback/dx.md` entry: `.mocharc.json`'s `node-option` array is dropped on Node 26 + mocha 11, so the documented `pnpm test -- --grep …` path fails any fixture with an `ssr` case.


LuLaValva:
Superseded by marko-js/htmljs-parser#239, which reports this from the parser's own tokenizer states rather than a source scan in the compiler. Marko picks it up through its existing `htmljs-parser` range once that publishes, so nothing is needed here.

PR #4007, opened 2026-08-19
docs(runtime-tags): note uncontrolled <let> is a seed, not a formula

DylanPiercey:
Uncontrolled `<let>` is deliberately a seed, not a formula: reactive initializers do not subscribe, so prefer `<const>` for derived values. Comments that at `setBindingDownstream(binding, false)` and drops the agent-feedback item that asked to warn on the shape.


DylanPiercey:
Closed: opened without an explicit ship call.

PR #4022, opened 2026-08-20
Error when a `load` import resolves to a class API tag

DylanPiercey:
A tags API template lazily importing a class API tag compiled to a dynamic tag referencing a binding whose import had already been removed, so `$setup` threw `ReferenceError: Child is not defined` at first render with no diagnostics. The compat dynamic tag path cannot read the load config, so the combination cannot work as written.

Reports it as a compile error at the import instead. The existing `lazy-class-child*` fixtures all use a class API parent and are unaffected.


DylanPiercey:
Fixed — the URL now points at `/docs/reference/lazy-loading`.

Worth noting it was broken rather than just suboptimal: `language.md` has no `#lazy-loading` anchor (that heading is "Lazy Tag `import`"), and it links out to the dedicated page itself.

PR #4052, opened 2026-08-21
fix(runtime-tags): don't reapply resumed scope payloads after a resume error

DylanPiercey:
A throw while processing resume data (e.g. a payload referencing a registry id missing under asset/deploy skew) skipped the consumption splice in `processResumes`, so every already-applied entry replayed on the next resume pass — reapplying scope payloads over newer state and permanently breaking later flushes.

Scope payloads are now consumed as they are applied (replaced with a no-op entry ahead of the invocation that can throw), so a later pass resumes the remaining entries exactly once; unconsumed effect strings still replay safely since their effects were discarded with the failed pass. Adds a fixture that streams a stale throwing payload between two await flushes and pins one application per entry plus recovery of interactivity. Size: +10 B min / +8 B brotli on the resume runtime.


DylanPiercey:
Closing: the motivating skew scenario is contrived (a page's html and hashed asset urls deploy together), a non-persisted resume throw has no recovery path regardless of post-throw list state, and persisted responses will carry a build id whose mismatch falls back to a full reload — so the hardening buys nothing at either layer.

PR #4054, opened 2026-08-21
refactor(runtime-tags): drop the reserved _d id, dispatching the dynamic tag script from its resume marker

DylanPiercey:
Removes the reserved `_d` registered effect id from the template id space — the motivation is the reserved-id-namespace cleanup from the agent-feedback item, not bytes: pages using branches grow ~20-30 B brotli (dynamic-tag pages ~50-95 B) while saving only ~5-20 B of wire per dynamic tag instance; bundles without branches are byte-identical.

The seam is a generic `visitEffects` table in resume keyed by visit symbol (mirroring `controllableScripts`): when a marker lands a branch and an entry exists for its symbol, resume pushes `(entry, branch)` into that pass's effects array, so the script runs under `runEffects` with every other resume effect; each pass also ticks the table's entries with no branch so a feature can retry work parked while its data was still streaming in a lazy channel (dropped when every channel has failed or drained). The dynamic tag feature installs the `BranchEndNativeTag` entry and owns all parking; the update and resume paths share one handler-attach helper keyed off the branch's root element. The server stops writing the per-instance `_d <branchId>` effect and the debug-only renderer scope prop.

The feedback item's other half — the placeholder dismiss (`_f`) — is deferred: a fully static async body flushes no markers or resume data at all, so dropping that id needs a swap-time signal from the inline reorder runtime. The item is rewritten to cover just that remainder.


DylanPiercey:
Closing after measurement: the visitEffects seam costs +179 B min / +92 B brotli of runtime on every page with branches (table + per-pass retry tick + parking array) against a ~10-15 B wire saving per handler-bearing dynamic native tag, so it only nets positive for a narrow page shape. The namespace-cleanup motivation also collapsed: `_e` is a value-identity serializer register consumed during data deserialization (can't be dispatched from a marker) and `_f` remains, so `meta.ts` and the `_`-prefix convention survive regardless. Keeping the reserved ids as-is; the trade-offs are recorded in the repo's feedback notes.

PR #4088, opened 2026-09-02
test(runtime-class): cover AsyncStream's node parsing and element stack

DylanPiercey:
`___getNode` is what turns a server-rendered stream into DOM nodes for insertion, and nothing exercised it: a single element comes back as that element, siblings are collected into a fragment, and the result is parsed once and cached. `parseHTML` reads the global document, so the block runs against one jsdom window and restores it afterwards.

Also covers the writer's element stack — a void element left open, an svg one self-closed, and nested begin/end pairs closing innermost first.

### DIFF
--- a/packages/runtime-class/test/components-browser/fixtures/template-mount/index.marko
+++ b/packages/runtime-class/test/components-browser/fixtures/template-mount/index.marko
@@ -1,0 +1,11 @@
+class {
+    onCreate() {
+        this.state = { label: "" };
+    }
+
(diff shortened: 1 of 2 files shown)


DylanPiercey:
Closing: the `___getNode` half covers html-mode-in-the-browser, which is not supported, and the element-stack half measured as zero coverage change (A/B: 55 units with and without). Following up with an audit of `___getNode` instead.

PR #4163, opened 2026-09-15
Fixture: dynamic tag bodies picked by input and by state

DylanPiercey:
Adds a fixture rendering two dynamic tags with the same child body, one picked by `input.tag` and one by a `<let>` a button toggles. It pins today's output, where both bodies serialize a scope and branch markers, so a follow-up that lets the input-picked body stay out of the payload shows as a snapshot diff.


DylanPiercey:
Folding the fixture into the improvement's PR.

